PR #1, opened 2024-06-25
feat: Implement slurm ops library

jedel1043:
Just a draft for now, but this is my idea of the library we can offer.

Still needs documentation + tests


lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -0,0 +1,110 @@
...
+
+    def install(self):
+        # TODO: Pin slurm to the stable channel
+        _snap("install", "slurm", "--channel", "edge")
+
+    def start(self):
+        _snap("start", "--enable", f"slurm.{self._service.value}")
+
+    def restart(self):
+        _snap("restart", f"slurm.{self._service.value}")
```
jedel1043, line 83: Thought: The original slurm manager uses different methods to restart slurm and munged, but maybe it would be better to have a single restart method for simplicity?

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -0,0 +1,110 @@
...
+        # TODO: use `slurm.setmungekey` when implemented
+        # subprocess.run(["slurm.setmungekey"], stdin=key)
+        key = base64.encodebytes(key)
+        _set_config(f"munge.key={key}")
+
+    def get_munge_key(self) -> bytes:
+        # TODO: use `slurm.setmungekey` when implemented
+        # key = subprocess.run(["slurm.getmungekey"])
+        key = _get_config("munge.key")
+        return base64.decodebytes(key.encode())
```
jedel1043, line 104: Do we also want to have a method to generate a munge key? What about the jwt key?

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043, line 1: Maybe the library could also handle the peer relation data to store the key between units? Though, I also see the appeal of keeping this juju-agnostic.

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: No, I would keep them separate. We shouldn't have to restart the Slurm services when cycling out the current munge.key for a new one - some clusters have a key refresh policy, especially when introducing new partitions - and we shouldn't have to restart munge when making updates to the Slurm configuration. It introduces more potential points of failure imo.

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: Good point - this is something that we should do.

The Slurm snap already generates a key at installation, and you would use `mungekey -c` to create a new key using the snap. We should have the snap generate the jwt, and create a new one when necessary.

What do you think about that?

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: I think it would be better to have that managed within a peer interface library that is embedded directly into the charm imo. slurmdbd, slurmrestd, and slurmctld all handle high-availability differently. 

slurmctld is the only charm where I'd see the need for us to have a peer relation to share the key between units. The rest would just fetch it using the requires relation they have with slurmctld.

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043: Yep, mirroring the munge key behaviour sounds good to me.

tests/integration/test_slurm_ops.py
NucciTheBoss, line 1: I think we can potentially style the integration tests for `slurm_ops` similar to the integration tests for the charm libraries in [`operator-libs-linux`](https://github.com/canonical/operator-libs-linux/blob/main/tests/integration/test_apt.py). Reason saying is that unlike `storage-libs`, we don't necessarily have any Juju-specifics in this lib such as relation data handling or custom events.

I'd rather see it such that we just spin up an LXD container, copy the test and necessary dependencies into the container, and then run the integration tests for `slurm_ops` inside the container. We could use [`gambol`](https://github.com/NucciTheBoss/gambol) for this.

That way we can focus more on the operations code itself. Let me know your thoughts on this!

NucciTheBoss (changes requested):
Looks good so far! Just a couple comments. Let me know if you have any questions!

tox.ini
```diff
@@ -10,7 +10,7 @@ min_version = 4.0.0
 [vars]
 src_path = {tox_root}/src
 tests_path = {tox_root}/tests
-;lib_path = {tox_root}/lib/charms/operator_name_with_underscores
+lib_path = {tox_root}/lib/charms/hpc_libs
```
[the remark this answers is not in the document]
jedel1043: Good catch!

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -0,0 +1,200 @@
...
+def _call(cmd: str, *args: [str]) -> bytes:
+    """Call a command with logging.
+
+    Raises:
+        subprocess.CalledProcessError: Raised if the command fails.
+    """
+    cmd = [cmd, *args]
+    _logger.debug(f"Executing command {cmd}")
+    try:
+        return subprocess.check_output(cmd, stderr=subprocess.PIPE, text=False)
```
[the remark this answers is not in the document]
NucciTheBoss: I suspected that was your reasoning; you would use `read_bytes(...)` if reading the munge key file directly in Python. 

Tbh I was thinking we make `getmungekey` and `setmungekey` take a base64 value directly and handle the decoding/encoding. That way all we have to worry about within the charms is just exchanging the string value back and forth via the relation data buckets. I personally find base64 strings much easier to work with compared to raw bytes objects :sweat_smile: 

However, we can go forward with this implementation you have here since we haven't updated the API for managing the munge key in the Slurm snap yet.

tests/integration/test_hpc_libs.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - name: "Install dependencies in a virtual environment"
+        run: |
+          export DEBIAN_FRONTEND=noninteractive
+          sudo apt update
+          sudo apt install -y python3-venv python3-yaml
+          python3 -m venv venv --system-site-packages
+          venv/bin/python3 -m pip install -r dev-requirements.txt
+      - name: "Run integration tests with pytest"
+        run: |
+          sudo venv/bin/python3 -m pytest -v \
```
NucciTheBoss, line 39: [nit]: I don't think you need `sudo` here as gambol playthroughs automatically run as root within the act instance, but I'll test it real quick myself. Not a huge deal; I'm just curious to see if `sudo` is actualy required. If it isn't required, I'll just yoink it by committing the suggestion below :smiley: 

```suggestion
      - name: "Install dependencies in a virtual environment"
        run: |
          export DEBIAN_FRONTEND=noninteractive
          apt update
          apt install -y python3-venv python3-yaml
          python3 -m venv venv --system-site-packages
          venv/bin/python3 -m pip install -r dev-requirements.txt
      - name: "Run integration tests with pytest"
        run: |
          venv/bin/python3 -m pytest -v \
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,52 @@
+# Copyright 2024 Canonical Ltd.
+# See LICENSE file for licensing details.
+
+name: cephfs-server-proxy tests
```
NucciTheBoss, line 4: Totally missed this on the first pass :rofl: - I'll commit the suggestion since you're enjoying the weekend!

```suggestion
name: hpc-libs tests
```

NucciTheBoss (approved):
Good stuff @jedel1043

I'm good with landing this as an initial implementation as we stabilize the API around the Slurm snap/deb :flight_arrival: 

Just two nits that I'm going to look into. Very small so I'll just commit these as suggestions since you're enjoying a well-earned weekend :star_struck:!

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -0,0 +1,200 @@
...
+"""
+
+import logging
+import functools
+import subprocess
+import yaml
+import base64
+import enum
+import tempfile
+import os
```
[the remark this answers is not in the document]
NucciTheBoss: _Now that the `lib_path` is the `ruff check ...` path, formatting is being applied_

tests/integration/test_hpc_libs.yaml
NucciTheBoss: Ran a test without using `sudo` and the test passed successfully. Doesn't look like we need to use `sudo` here :smile:

PR #4, opened 2024-07-03
feat: add workflow for publishing libraries after merge

NucciTheBoss:
Title gives it away, but this pull request sets up the `release-libs` workflow for `hpc-libs`. Pretty much the same thing as the the workflow from `storage-libs`.

I added a token to the repository secrets that will allow us to upload the `slurm_ops` library. It will expire while we are in The Hague for the Engineering Sprint in October. I'm very tempted to write us a tool like [`tokenator`](https://github.com/snapcrafters/tokenator) that enables us to bulk refresh our repo secrets so it's not this weird song and dance when we try to publish new updates to Charmhub :sweat_smile: 

### Misc.

I added a _.editorconfig_ file. Helps with consistency across IDE's. We can extend rules farther as we need to :smiley:

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,15 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+
+[*.{py}]
+indent_style = space
+indent_size = 4
+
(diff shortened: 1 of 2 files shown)


jedel1043 (approved):
LGTM!

PR #6, opened 2024-07-04
feat: add `set_configs` operation to `SlurmManager`

jedel1043:
Implements a simple utility method to be able to set configs in bulk. This is required by the Slurm charms since they generate and set configs in bulk.

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -57,6 +57,7 @@ import logging
 import os
 import subprocess
 import tempfile
+from collections.abc import Mapping
 
 import yaml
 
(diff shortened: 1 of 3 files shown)


NucciTheBoss (approved):
LGTM :rocket:

NucciTheBoss:
@jedel1043 overall, I'm good with the implementation, but I have an idea that I'd like to know your thoughts on.

Why not have `set_config(...)` just accept a mapping rather than two positional arguments for both key and value? That way we don't need to have two methods that somewhat implement the same functionality. Reason saying is that when I look at a method like `set_config(...)`, I think of `snapctl set ...` which accepts multiple configuration arguments.

I think it could potentially be better to have just one `set_config(...)` method that takes in a mapping of multiple configuration options. I see us using `set_configs(...)` way more often than `set_config(...)` as we'll likely need to handle a variable amount of configuration options as they come in from relations, actions, etc. rather than granular, single option changes.

Let me know what your thoughts are :smile:

jedel1043:
The main reason is to avoid breaking changes, but if we wanna do them, I think your suggestion would be better, yeah.

NucciTheBoss:
> The main reason is to avoid breaking changes, but if we wanna do them, I think your suggestion would be better, yeah.

You have the right idea, but I don't think we have to worry about breaking changes right now given that this lib is very much in development, and we'll very much likely be the only consumers. Better to get it how we want it now rather than later when we're _stable_ :sweat_smile:

Let's go ahead with the proposed approach of having `set_config(...)` accept a mapping rather than two positional arguments so we only have one method for getting configuration options.

NucciTheBoss:
#8 adds the `set_config(...)` functionality where you're able to set multiple Slurm configurations at once.

jedel1043:
Superseeded by that then.

PR #9, opened 2024-07-10
Add `is_container` charm library

NucciTheBoss:
This pull request adds the small `is_container` charm library. This utility enables us to check if the current charm is running within a container using the [`systemd-detect-virt`](https://manpages.ubuntu.com/manpages/noble/en/man1/systemd-detect-virt.1.html) utility.

Why add this library? In multiple places we're using a utility like this - like both the Slurm and NFS operators - so it would be good to just have one source of truth for this method. We can also extend as necessary to detect multiple different container runtimes.

### DIFF
--- a/lib/charms/hpc_libs/v0/is_container.py
+++ b/lib/charms/hpc_libs/v0/is_container.py
@@ -54,8 +54,8 @@ LIBAPI = 0
 LIBPATCH = 1
 
 
-class DetectVirtNotFoundError(Exception):
-    """Raise error if `systemd-detect-virt` executable is not found on machine."""
+class UnknownVirtStateError(Exception):
+    """Raise error if unknown virtualization state is returned."""
 
     @property
     def message(self) -> str:
@@ -70,7 +70,7 @@ def is_container() -> bool:
         DetectVirtNotFoundError: Raised if `systemd-detect-virt` is not found on machine.
     """
     if shutil.which("systemd-detect-virt") is None:
-        raise DetectVirtNotFoundError(
+        raise UnknownVirtStateError(
             (
                 "executable `systemd-detect-virt` not found. "
                 + "cannot determine if machine is a container instance"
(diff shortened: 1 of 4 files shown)


lib/charms/hpc_libs/v0/is_container.py
```diff
@@ -0,0 +1,85 @@
...
+"""Detect if machine is a container instance.
+
+Even though Juju supports using LXD containers as the backing cloud for
+deploying charmed operators, not all HPC applications work within system containers,
+and some need additional configuration. This simple charm library provides the `is_container`
+function, a simple, deterministic way to identify if the charm is deployed into
+a system container using `systemd-detect-virt`.
+
+Exit code 0 means that the charm is running with a container. A non-zero exit code
+means that the charm is not running within a container.
```
jedel1043, line 24: Maybe we could skip mentioning the internals of the library? It would be better so that we don't need to change the documentation if the implementation changes.

lib/charms/hpc_libs/v0/is_container.py
```diff
@@ -0,0 +1,85 @@
...
+LIBPATCH = 1
+
+
+class DetectVirtNotFoundError(Exception):
+    """Raise error if `systemd-detect-virt` executable is not found on machine."""
+
+    @property
+    def message(self) -> str:
+        """Return message passed as argument to exception."""
+        return self.args[0]
```
jedel1043, line 67: I think this error should be more generic, something like `UnknownVirtState`. We can add specifics about how it didn't find `systemd-detect-virt` in the message, but a more generic error would help if we add more ways to detect virtualization in the future.

jedel1043 (commented):
Overall looks good! I just have some suggestions and discussion points.

lib/charms/hpc_libs/v0/is_container.py
NucciTheBoss: Sure sure, I can drop it. `returncode == 0` already implies that an exit code of 0 means that we are running inside of a container.

lib/charms/hpc_libs/v0/is_container.py
NucciTheBoss: Good idea. I'll make this change then.

jedel1043 (approved):
Looks perfect to me!

PR #12, opened 2024-07-10
feat: enable bulk getting and unsetting of configuration options

NucciTheBoss:
This pull request adds the ability to both bulk get and unset configuration options on slurm config objects. This functionality will be helpful for both if we want to reset the state of Slurm, and quickly get configuration information for various Slurm objects for comparison against new information received by the charms.

For example:

```python3
slurmctld = SlurmManagerBase(ServiceType.SLURMCTLD)

# Get the entire Slurm controller configuration
slurmctld.config.get()

# Reset the configuration
slurmctld.config.unset()
```

### Misc.

* Fixes a bug with `.unset(...)`. If we use `snap unset ...` instead of `snap set ...`, we don't need to append the `!` exclamation point to the end of the key name. Whoopsies :cold_face:
* `ConfigurationManager` now takes a string in its constructor rather than a `ServiceType`. This makes it much easier to create configuration managers for slurmctld components such as `cgroup` and `acct_gather`. I realized that `ConfigurationManager` only needs `config_name`, so there's no need to actually pass the full `ServiceType` enum.

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -59,6 +59,7 @@ __all__ = [
     "format_key",
     "install",
     "version",
+    "ConfigurationManager",
     "ServiceType",
     "SlurmManagerBase",
 ]
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -202,8 +202,8 @@ class ServiceManager:
 class ConfigurationManager:
     """Control configuration of a Slurm component."""
 
-    def __init__(self, service: ServiceType) -> None:
-        self._service = service
+    def __init__(self, name: str) -> None:
+        self._name = name
 
     def get_options(self, *keys: str) -> Mapping[str, Any]:
         """Get given configurations values for Slurm component."""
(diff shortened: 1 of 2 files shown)


NucciTheBoss:
> I'll bump the LIBPATCH version after we discuss & approve

jedel1043 (commented):
Agreed with all the changes, it makes it really easy to get/unset all the slurm config at once.
However, we were planning on adding some default config to fix some slurm limitations for the snap (https://github.com/charmed-hpc/slurm-snap/issues/22), and I worry that using `unset` to clear everything could become a pitfall.

NucciTheBoss:
Hmm... you make a point. Currently, `slurmctld-operator` defines the base set of configuration options it needs, so at least in the charm context we can just go back to that base state:

```python3
slurmctld = SlurmManagerBase(ServiceType.SLURMCTLD)
slurmctld.config.unset()
slurmctld.set_defaults()
```

From the snap context, if the entire configuration is blown out we could just add some logic to the `configure` hook that's like "if `snap get -d slurm <option>` is empty, revert to these defaults". How does that sound?

jedel1043:
> From the snap context, if the entire configuration is blown out we could just add some logic to the `configure` hook that's like "if `snap get -d slurm <option>` is empty, revert to these defaults". How does that sound?

Yep, sounds good!

jedel1043 (approved):

PR #13, opened 2024-07-10
feat: Add `SlurmOpsError` to `slurm_ops`

### LINKED ISSUE #11: [Enhancement]: Return library error on `slurm_ops` instead of propagating `CalledProcessError`
Should be a pretty nice enhancement, considering that returning `CalledProcessError` is a bit more opaque to the user in contrast to having a nifty `SlurmOpsError` that is visible from the module exports.

jedel1043:
Fixes #11.

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -91,6 +91,15 @@ _acronym = re.compile(r"(?<=[A-Z])(?=[A-Z][a-z])")
 _kebabize = re.compile(r"(?<=[a-z0-9])(?=[A-Z])")
 
 
+class SlurmOpsError(Exception):
+    """Exception raised when a slurm operation failed."""
+
+    @property
+    def message(self) -> str:
+        """Return message passed as argument to exception."""
+        return self.args[0]
+
+
 def format_key(key: str) -> str:
     """Format Slurm configuration keys from SlurmCASe into kebab case.
 
@@ -129,7 +138,7 @@ def _call(cmd: str, *args: str, stdin: Optional[str] = None) -> str:
     """Call a command with logging.
 
     Raises:
-        subprocess.CalledProcessError: Raised if the command fails.
+        SlurmOpsError: Raised if the command fails.
     """
     cmd = [cmd, *args]
     _logger.debug(f"Executing command {cmd}")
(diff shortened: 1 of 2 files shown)


lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -129,7 +137,7 @@ def _call(cmd: str, *args: str, stdin: Optional[str] = None) -> str:
     """Call a command with logging.
 
     Raises:
-        subprocess.CalledProcessError: Raised if the command fails.
+        SlurmManagerError: Raised if the command fails.
```
NucciTheBoss, line 140: I think you mean `SlurmOpsError` :smile: 

```suggestion
        SlurmOpsError: Raised if the command fails.
```

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -91,6 +91,14 @@ def _on_install(self, _) -> None:
 _kebabize = re.compile(r"(?<=[a-z0-9])(?=[A-Z])")
 
 
+class SlurmOpsError(BaseException):
```
NucciTheBoss, line 94: Why `BaseException` instead of `Exception`. [This documentation piece from Python](https://docs.python.org/3/library/exceptions.html) seems to indicate that we should derive our custom exception from `Exception` instead. 

Tbh we could copy the exception from `is_container` which just returns the first arg passed to the exception as the message.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -138,7 +146,7 @@ def _call(cmd: str, *args: str, stdin: Optional[str] = None) -> str:
     except subprocess.CalledProcessError as e:
         _logger.error(f"`{' '.join(cmd)}` failed")
         _logger.error(f"stderr: {e.stderr.decode()}")
-        raise
+        raise SlurmOpsError(f"Command {cmd[0]} failed. Reason:\n{e.stderr.decode()}")
```
NucciTheBoss, line 149: [nit]: should all error messages be lowercase or should they be titlelized?

At least in Go, the strong preference is that all error messages should be lowercase. What are your thoughts on this :eyes:?

NucciTheBoss (changes requested):
Just a couple comments but overall LGTM :+1:

It will be good to have a custom error message that we can handle.

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043: Oops

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043: lowercase seems better for nice displays: `Error: important message` vs `Error: Important message`

jedel1043:
Won't bump the version so that we can bump in #12 instead and avoid publishing two consecutive versions.

NucciTheBoss (approved):
Ship it :ship: 

_I'll bump the LIBPATCH version in my pull request after rebase. That way we only need to publish one update to Charmhub._

PR #17, opened 2024-07-12
feat: add method to get the host name

### LINKED ISSUE #16: [Enhancement]: Add `hostname` method to `SlurmManagerBase`
One thing that I noticed while working with `SlurmManagerBase` is that all the Slurm charms need a way to retrieve their hostname to share with the other Slurm daemons. We should add a hostname method so that all the Slurm service managers have a common way for retrieving the unit hostname.

Not a difficult thing to implement!

```python3
import socket

socket.gethostname().split(".")[0]
```

jedel1043:
Closes #16.

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -67,6 +67,7 @@ __all__ = [
 import json
 import logging
 import re
+import socket
 import subprocess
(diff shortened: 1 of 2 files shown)


jedel1043:
Should be merged after #15 to avoid bumping the version number twice.

NucciTheBoss (approved):
LGTM! Ship it :ship:

PR #20, opened 2024-07-16
feat: add prometheus exporter manager

### LINKED ISSUE #19: [Enhancement]: Add component to manage prometheus exporter
By default, the Slurm prometheus exporter comes disabled with the Slurm snap. It makes sense to have the exporter disabled on installation because:

a. because Slurm needs further configuration out of the box before it is ready to be started by the administrator.
b. An operator may choose to not deploy COS.

We should have a component that allows us to control the exporter from inside the Slurm charms. That way we can turn it on once COS has been integrated into a Slurm cluster deployment.

NucciTheBoss:
Adds the `PrometheusExporterManager` class for managing the operations of the Slurm prometheus exporter bundled with the Slurm snap.

`PrometheusExporterManager` inherits from `ServiceManager` so that it can enable, disable, and restart the exporter service. The exporter is exposed through the `exporter` attribute of `SlurmManagerBase`. This way we can use something like the following block of code to turn on the exporter after we have connected a COS deployment to Charmed HPC:

```python3
def _on_integration(self, _: Event) -> None:
    """Handler for when COS is integrated into cluster."""
    # staging operations for setting up observability.
    ...
    self._manager.exporter.enable()
```

One thing we'll likely need to do in the future is that once we determine how we want to configure the prometheus exporter, we'll need to expose a config interface.

Fixes #19

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -84,7 +84,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 4
+LIBPATCH = 5
 
 # Charm library dependencies to fetch during `charmcraft pack`.
 PYDEPS = ["pyyaml>=6.0.1"]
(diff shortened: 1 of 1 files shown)


jedel1043 (approved):
Nice! This will make it really easy to integrate COS into all the charms.

NucciTheBoss:
Let's go! I'll bump the LIBPATCH version and then merge.

PR #21, opened 2024-07-26
fix: remove `decode` inside `_call`

jedel1043:
Fixes a bug where the script would throw when `check_output` throws an error, since the returned `stderr` would be a str instead of bytes.


NucciTheBoss (approved):
LGTM! :tada:

PR #22, opened 2024-08-19
feat: abstract around base package manager

jedel1043:
This PR abstracts around the base package manager used.

The objective of this is to enable switching between snap packages and deb packages in a more composable way.

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -26,14 +26,13 @@ The `SlurmManagerBase` constructor receives a `ServiceType` enum. The enum instr
 the inheriting Slurm service manager how to manage its corresponding Slurm service on the host.
 
 ```python3
-import charms.hpc_libs.v0.slurm_ops as slurm
-from charms.hpc_libs.v0.slurm_ops import SlurmManagerBase, ServiceType
+from charms.hpc_libs.v0.slurm_ops import SlurmManagerBase, ServiceType,
 
 class SlurmctldManager(SlurmManagerBase):
     # Manage `slurmctld` service on host.
 
     def __init__(self) -> None:
-        super().__init__(ServiceType.SLURMCTLD)
+        super().__init__(ServiceType.SLURMCTLD, snap=True)
 
 
 class ApplicationCharm(CharmBase):
@@ -49,32 +48,41 @@ class ApplicationCharm(CharmBase):
         )
 
     def _on_install(self, _) -> None:
-        slurm.install()
-        self.unit.set_workload_version(slurm.version())
+        self._slurm_manager.install()
+        self.unit.set_workload_version(self._slurm_manager.version())
         self._slurm_manager.config.set({"cluster-name": "cluster"})
 ```
 """
 
 __all__ = [
-    "format_key",
-    "install",
-    "version",
-    "ConfigurationManager",
-    "ServiceType",
-    "SlurmManagerBase",
     "SlurmOpsError",
+    "ServiceType",
+    "SlurmOpsManager",
+    "ServiceManager",
+    "MungeKeyManager",
+    "MungeManager",
+    "SnapManager",
+    "SlurmManagerBase",
+    "SlurmctldManager",
+    "SlurmdManager",
+    "SlurmdbdManager",
+    "SlurmrestdManager",
 ]
 
-import json
 import logging
-import re
+import os
 import socket
 import subprocess
+from abc import ABC, abstractmethod
 from collections.abc import Mapping
+from contextlib import contextmanager
 from enum import Enum
-from typing import Any, Optional
+from pathlib import Path
+from typing import Any, Optional, Union
 
+import dotenv
 import yaml
+from slurmutils.editors import slurmconfig, slurmdbdconfig
 
 # The unique Charmhub library identifier, never change it
 LIBID = "541fd767f90b40539cf7cd6e7db8fabf"
@@ -84,58 +92,12 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 5
+LIBPATCH = 6
 
 # Charm library dependencies to fetch during `charmcraft pack`.
-PYDEPS = ["pyyaml>=6.0.1"]
+PYDEPS = ["pyyaml>=6.0.2", "python-dotenv~=1.0.1", "slurmutils~=0.6.0"]
 
 _logger = logging.getLogger(__name__)
-_acronym = re.compile(r"(?<=[A-Z])(?=[A-Z][a-z])")
-_kebabize = re.compile(r"(?<=[a-z0-9])(?=[A-Z])")
-
-
-class SlurmOpsError(Exception):
-    """Exception raised when a slurm operation failed."""
-
-    @property
-    def message(self) -> str:
-        """Return message passed as argument to exception."""
-        return self.args[0]
-
-
-def format_key(key: str) -> str:
-    """Format Slurm configuration keys from SlurmCASe into kebab case.
-
-    Args:
-        key: Slurm configuration key to convert to kebab case.
-
-    Notes:
-       Slurm configuration syntax does not follow proper PascalCasing
-       format, so we cannot put keys directly through a kebab case converter
-       to get the desired format. Some additional processing is needed for
-       certain keys before the key can properly kebabized.
-
-       For example, without additional preprocessing, the key `CPUs` will
-       become `cp-us` if put through a kebabizer with being preformatted to `Cpus`.
-    """
-    if "CPUs" in key:
-        key = key.replace("CPUs", "Cpus")
-    key = _acronym.sub(r"-", key)
-    return _kebabize.sub(r"-", key).lower()
-
-
-def install() -> None:
-    """Install Slurm."""
-    # FIXME: Pin slurm to the stable channel
-    _snap("install", "slurm", "--channel", "latest/candidate", "--classic")
-
-
-def version() -> str:
-    """Get the current version of Slurm installed on the system."""
-    info = yaml.safe_load(_snap("info", "slurm"))
-    if (ver := info.get("installed")) is None:
-        raise SlurmOpsError("unable to retrive snap info. Ensure slurm is correctly installed")
-    return ver.split(maxsplit=1)[0]
 
 
 def _call(cmd: str, *args: str, stdin: Optional[str] = None) -> str:
@@ -151,7 +113,7 @@ def _call(cmd: str, *args: str, stdin: Optional[str] = None) -> str:
     except subprocess.CalledProcessError as e:
         _logger.error(f"`{' '.join(cmd)}` failed")
         _logger.error(f"stderr: {e.stderr.decode()}")
-        raise SlurmOpsError(f"command {cmd[0]} failed. Reason:\n{e.stderr.decode()}")
+        raise SlurmOpsError(f"command {cmd[0]} failed. reason:\n{e.stderr.decode()}")
 
 
 def _snap(*args) -> str:
@@ -163,17 +125,13 @@ def _snap(*args) -> str:
     return _call("snap", *args)
 
 
-def _mungectl(*args: str, stdin: Optional[str] = None) -> str:
-    """Control munge via `slurm.mungectl ...`.
+class SlurmOpsError(Exception):
+    """Exception raised when a slurm operation failed."""
 
-    Args:
-        *args: Arguments to pass to `mungectl`.
-        stdin: Input to pass to `mungectl` via stdin.
-
-    Raises:
-        subprocess.CalledProcessError: Raised if `mungectl` command fails.
-    """
-    return _call("slurm.mungectl", *args, stdin=stdin)
+    @property
+    def message(self) -> str:
+        """Return message passed as argument to exception."""
+        return self.args[0]
 
 
 class ServiceType(Enum):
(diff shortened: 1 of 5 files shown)


NucciTheBoss:
I think we're on the right path here with enabling support for Slurm in both snap and deb form, but I'm wondering where we can "duplicate" things between package formats such that we don't need to have as much abstraction happening in the background to manipulate things like the munge key file and Slurm configuration in the background.

Re. the Slurm configuration, we can just manipulate that with `slurmutils`. For the munge key, we need to enable support for changing the key file location in `mungectl`.

jedel1043:
@NucciTheBoss Yeah, using mungectl and slurmutils on both snap and deb mode would be nice. In the meantime, I think we can just push the current abstractions and see if we can unify things later.

jedel1043:
Right, this won't pass CI until we fix https://github.com/charmed-hpc/slurm-snap/issues/47

NucciTheBoss:
> Right, this won't pass CI until we fix [charmed-hpc/slurm-snap#47](https://github.com/charmed-hpc/slurm-snap/issues/47)

Yeah... I think it's time for `slurmhelpers` to go :disappointed:

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -197,9 +148,183 @@ def config_name(self) -> str:
...
+        self.exporter = PrometheusExporterManager(ops_manager)
+        self.ops = ops_manager
+
+    @property
+    def hostname(self) -> str:
+        """The hostname where this manager is running."""
+        return socket.gethostname().split(".")[0]
+
+
+class SnapPackageManager(SlurmPackageManager):
```
NucciTheBoss, line 294: Should this just be `SnapManager` instead? Reason saying is that it's also providing methods for accessing both the service and configuration managers. _Less characters to type the better_ :rofl:

```suggestion
class SnapManager(SlurmPackageManager):
```

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -197,9 +148,183 @@ def config_name(self) -> str:
...
+        """Get a `ServiceManager` for the specified `service`."""
+        return _SnapServiceManager(service)
+
+    def configuration_manager_for(self, service: ServiceType) -> ConfigurationManager:
+        """Get a `ConfigurationManager` for the specified `service`."""
+        return _SnapConfigurationManager(service.config_name)
+
+    def mungekey_manager(self) -> MungekeyManager:
+        """Get the `MungekeyManager` for munge."""
+        return _SnapMungekeyManager()
```
NucciTheBoss, line 319: Tbh - I'm wondering if these should be properties instead? They're just providing proxied access to the objects that provide the specific methods for the given packaging format.

```suggestion
    @property
    def service_manager(self, service: ServiceType) -> ServiceManager:
        """Get a `ServiceManager` for the specified `service`."""
        return _SnapServiceManager(service)

    @property
    def config_manager(self, service: ServiceType) -> ConfigurationManager:
        """Get a `ConfigurationManager` for the specified `service`."""
        return _SnapConfigurationManager(service.config_name)

    @property
    def key_manager(self) -> MungekeyManager:
        """Get the `MungekeyManager` for munge."""
        return _SnapMungekeyManager()
```

Let me know your thoughts here :smile:

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -223,23 +348,17 @@ def active(self) -> bool:
...
         return "inactive" not in services[f"slurm.{self._service.value}"]
 
+    def type(self) -> ServiceType:
+        """Return the service type of the managed service."""
+        return self._service
 
-class ConfigurationManager:
-    """Control configuration of a Slurm component."""
+
+class _SnapConfigurationManager(ConfigurationManager):
```
NucciTheBoss, line 356: `_SnapConfigManager` :pleading_face::point_right::point_left:

 ```suggestion
class _SnapConfigManager(ConfigurationManager):
```

File: lib/charms/hpc_libs/v0/slurm_ops.py [hunk 4]
```diff
@@ -197,9 +148,183 @@ def config_name(self) -> str:
...
+
+class SlurmManagerBase:
+    """Base manager for Slurm services."""
+
+    def __init__(self, service: ServiceType, ops_manager: SlurmPackageManager) -> None:
+        self.service = ops_manager.service_for(service)
+        self.config = ops_manager.configuration_manager_for(service)
+        self.munge = MungeManager(ops_manager)
+        self.exporter = PrometheusExporterManager(ops_manager)
+        self.ops = ops_manager
```
NucciTheBoss, line 286: I'm having a thought here :monocle_face: 

What if we used a boolean flag here to indicate which operations manager we're using rather than needing to directly pass a package manager object? We know that we're only going to support deb and snap, so it shouldn't be too complicated to implement:

```python3
def __init__(self, service: ServiceType, snap: bool = False) -> None:
    ops_manager = SnapPackageManager() if snap else DebPackageManager()
    self.service = ops_manager.service_manager
    self.config = ops_manager.config_manager
    self.munge = ops_manager.munge_manager
    self.exporter = ops_manager.exporter_manager
    self.install = ops_manager.install
    self.version = ops_manager.version
```

This way we can have the default behavior be to use the Debian package, change snap to True when we're ready to switch over, and we can bind the operations manager methods and properties to where we need them to be. We also don't have to expose the methods for retrieving the format-specific operations managers. We also don't need to have the `ops` attribute then to expose the `install` and `version` methods and can just call them the way that we did previously.

Let me know your thoughts on this!

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -257,52 +376,62 @@ def unset(self, *keys: str) -> None:
...
+        SlurmOpsError: Raised if the command fails.
+    """
+    cmd = [cmd, *args]
+    _logger.debug(f"Executing command {cmd}")
+    try:
+        return subprocess.check_output(cmd, input=stdin, stderr=subprocess.PIPE, text=True).strip()
+    except subprocess.CalledProcessError as e:
+        _logger.error(f"`{' '.join(cmd)}` failed")
+        _logger.error(f"stderr: {e.stderr.decode()}")
+        raise SlurmOpsError(f"command {cmd[0]} failed. Reason:\n{e.stderr.decode()}")
```
NucciTheBoss, line 428: [nit]: I wonder if we should just have the entire error message be lowercase? 

```suggestion
        raise SlurmOpsError(f"command {cmd[0]} failed. reason:\n{e.stderr.decode()}")
```

NucciTheBoss (changes requested):
Great steps in the right direction! I left a couple comments that I would like to know your thoughts on.

Big thing I'm wondering is how we can potentially simplify the creation of the `SlurmManagerBase` based on which package format we are using.

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043: Good idea! That's both simpler for us, and it could also be extended later if we do have to support an additional package manager.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -240,6 +192,175 @@ def get_options(self, *keys: str) -> Mapping[str, Any]:
...
+
+    def set(self, key: str) -> None:
+        """Set a new munge key.
+
+        Args:
+            key: A new, base64-encoded munge key.
+        """
+
+    def generate(self) -> None:
+        """Generate a new, cryptographically secure munge key."""
```
NucciTheBoss, line 226: Since this is a abstract class, shouldn't these methods be decorated with the `abstractmethod` decorator? That way Python will complain if we don't implement the `get`, `set`, and `generate` methods on inheriting objects:

```suggestion
class MungeKeyManager(ABC):
    """Control the munge key."""

    @abstractmethod
    def get(self) -> str:
        """Get the current munge key.

        Returns:
            The current munge key as a base64-encoded string.
        """

    @abstractmethod
    def set(self, key: str) -> None:
        """Set a new munge key.

        Args:
            key: A new, base64-encoded munge key.
        """
   
   @abstractmethod
    def generate(self) -> None:
        """Generate a new, cryptographically secure munge key."""
```

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -49,30 +48,34 @@ def __init__(self, *args, **kwargs) -> None:
         )
 
     def _on_install(self, _) -> None:
-        slurm.install()
-        self.unit.set_workload_version(slurm.version())
+        self._slurm_manager.ops.install()
+        self.unit.set_workload_version(self._slurm_manager.ops.version())
```
NucciTheBoss, line 52: We can drop the `ops` property here now since we changed how the manager object is constructed now:

```suggestion
        self._slurm_manager.install()
        self.unit.set_workload_version(self._slurm_manager.version())
```

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -26,14 +26,13 @@
...
-import charms.hpc_libs.v0.slurm_ops as slurm
-from charms.hpc_libs.v0.slurm_ops import SlurmManagerBase, ServiceType
+from charms.hpc_libs.v0.slurm_ops import SlurmManagerBase, ServiceType, SnapManager,
 
 class SlurmctldManager(SlurmManagerBase):
     # Manage `slurmctld` service on host.
 
     def __init__(self) -> None:
-        super().__init__(ServiceType.SLURMCTLD)
+        super().__init__(ServiceType.SLURMCTLD, SnapManager())
```
NucciTheBoss, line 35: We be booling:

```suggestion
        super().__init__(ServiceType.SLURMCTLD, snap=True)
```

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -26,14 +26,13 @@
 the inheriting Slurm service manager how to manage its corresponding Slurm service on the host.
 
 ```python3
-import charms.hpc_libs.v0.slurm_ops as slurm
-from charms.hpc_libs.v0.slurm_ops import SlurmManagerBase, ServiceType
+from charms.hpc_libs.v0.slurm_ops import SlurmManagerBase, ServiceType, SnapManager,
```
NucciTheBoss, line 29: IIRC we don't need SnapManager here now:

```suggestion
from charms.hpc_libs.v0.slurm_ops import SlurmManagerBase, ServiceType
```

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -240,6 +192,175 @@ def get_options(self, *keys: str) -> Mapping[str, Any]:
...
+    def install(self) -> None:
+        """Install Slurm using the `slurm` snap."""
+        # FIXME: Pin slurm to the stable channel
+        _snap("install", "slurm", "--channel", "latest/candidate", "--classic")
+
+    def version(self) -> str:
+        """Get the current version of the `slurm` snap installed on the system."""
+        info = yaml.safe_load(_snap("info", "slurm"))
+        if (ver := info.get("installed")) is None:
+            raise SlurmOpsError("unable to retrive snap info. Ensure slurm is correctly installed")
```
NucciTheBoss, line 304:
```suggestion
            raise SlurmOpsError("unable to retrive snap info. ensure slurm is correctly installed")
```

NucciTheBoss (changes requested):
This is looking really good! Just a couple comments, but I like how clean this implementation is looking!

jedel1043:
@NucciTheBoss We need to integrate `slurmutils` to make CI pass 😅
Currently working on the `update` operation for slurmutils.

jedel1043:
Wow... CI passed...

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -197,9 +180,244 @@ def config_name(self) -> str:
...
+        """Get the config manager of slurmctld."""
+        with slurmdbdconfig.edit(self._config_path) as config:
+            yield config
+
+
+class SlurmrestdManager(SlurmManagerBase):
+    """Manager for the Slurmrestd service."""
+
+    def __init__(self, *args, **kwargs) -> None:
+        super().__init__(service=ServiceType.SLURMRESTD, *args, **kwargs)
```
NucciTheBoss, line 412: Should we instead have these defined in their respective charms instead? 

I know we discussed earlier about having these custom methods for managing configuration of the services, but what I'm concerned about is now that we have these defined here, anytime we update one manager, we end up needing to update the lib across all the charms.

This way we can also discuss the implementation of each manager in its own PR against the relevant charm.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -163,17 +129,34 @@ def _snap(*args) -> str:
     return _call("snap", *args)
 
 
-def _mungectl(*args: str, stdin: Optional[str] = None) -> str:
-    """Control munge via `slurm.mungectl ...`.
+def format_key(key: str) -> str:
```
NucciTheBoss, line 132: Are we still using this method anywhere? Quick search for usages seems to indicate that we're no longer using it except within the unit tests. Side effect of not letting the snap dictate everything.

NucciTheBoss (changes requested):
I'm quite happy with how this has shaped up. Just a couple suggested changes and I'm ready to merge this so we can get focused on Terraform x AMD GPU stuff :star_struck: 

1. I don't think we're using the `format_key` function anymore since we're editing the config files directly.
2. I think we can worry about linting the passed environment variables later when we have a better idea of where inputs are coming from within the Slurm charms. I think right now we just set/get/unset with no frills about it.
3. Should we define the Slurm service managers in their own charms respectively? I'm thinking about that discussion we had a while ago where we didn't want to put the managers into this lib because then we'd need to update the lib everywhere everytime we made a change to the manager.

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043: I like the current design because it abstracts away the interactions with `OpsManager`, and hides everything behind a `snap: bool` flag. Moving this to the charms would do the opposite, forcing us to use `OpsManager` from the charms.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -197,9 +180,244 @@ def config_name(self) -> str:
...
+    """Control configuration of environment variables used in Slurm components."""
+
+    def __init__(self, file: Union[str, os.PathLike], prefix: str, keys: [str]) -> None:
+        self._file: Path = Path(file)
+        self._service = prefix
+        self._keys: [str] = keys
+
+    def _config_to_env_var(self, key: str) -> str:
+        """Get the environment variable corresponding to the configuration `key`."""
+        return self._service.replace("-", "_").upper() + "_" + key.replace("-", "_").upper()
```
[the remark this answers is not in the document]
jedel1043: This wasn't to preserve Snap syntax, it was to preserve the "official" name for the slurmd option `config-server`, which I'm not sure if we should change.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -197,9 +180,244 @@ def config_name(self) -> str:
...
+        """Generate a new, cryptographically secure munge key."""
+
+
+class _EnvManager:
+    """Control configuration of environment variables used in Slurm components."""
+
+    def __init__(self, file: Union[str, os.PathLike], prefix: str, keys: [str]) -> None:
+        self._file: Path = Path(file)
+        self._service = prefix
+        self._keys: [str] = keys
```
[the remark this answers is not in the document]
jedel1043: I added this because I needed it 😅
Previously I tried to set some env variable for the charms (I think it was for the Mysql issue), but it didn't really work, and after minutes of debugging I just realized I had made a typo on the variable name 🤣

The good thing about doing the sanity check is that it makes silly errors like this noticeable, and not silently do the wrong thing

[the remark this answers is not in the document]
jedel1043: Though, If we're exposing the configs as properties instead of exposing the env manager directly, I think we can safely remove the sanity checks, since the errors will be caught by our tests anyways.

lib/charms/hpc_libs/v0/slurm_ops.py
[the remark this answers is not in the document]
jedel1043: > but we need to make sure that we override the various service files to use the prefixed variable names.

Yep. I'm thinking that the `install` hook for the deb manager should be able to do this automatically

lib/charms/hpc_libs/v0/slurm_ops.py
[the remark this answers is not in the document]
jedel1043: Removed the checks. Any errors should be caught by unit tests anyways.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -197,9 +180,244 @@ def config_name(self) -> str:
...
+
+    @max_thread_count.setter
+    def max_thread_count(self, count: int) -> None:
+        """Set the max number of threads that munged can use."""
+        self._env_manager.set({"max-thread-count": count})
+
+    @max_thread_count.deleter
+    def max_thread_count(self, count: int) -> None:
+        """Unset the max number of threads that munged can use."""
+        self._env_manager.unset("max-thread-count")
```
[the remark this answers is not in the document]
jedel1043: Removed for now. We can add more configs later if we need to.

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: Okay, sounds good to me then. Having `config` exposed as a context manager, we can just update the the current configuration using the `update(...)` method. 

We'll still want to inherit from the dedicated service managers in their respective charms, correct? We'll still need to compose certain functionality on `SlurmctldManager` like the ability to edit the cgroup or OCI image configuration. Are you envisioning that this functionality would be defined within `hpc_libs`, or would you see this as something we add within the slurmctld charm?

jedel1043: I'm envisioning that any additional functionality would start on a "testing phase" in the deb version of the charm, and if we determine that the functionality is good enough and also useful for the snap version, we would move it to `hpc_libs`

lib/charms/hpc_libs/v0/slurm_ops.py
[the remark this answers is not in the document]
NucciTheBoss: > it was to preserve the "official" name for the slurmd option

Where's this official name defined? I know that `config-server` is the name we use for the option on the snap configuration, but the CLI flag is `--conf-server`. IIRC the deb version of slurmd just uses the environment variables `SLURMD_OPTIONS`, so I don't think having it as `config-server` is a hard and fast rule.

TBH we could just have the option map to the environment variable `SLURMD_CONFIG_SERVER`, and let the snap and deb worry about how to interpret the env file.

NucciTheBoss:
This is looking great @jedel1043 :tada: 

I just replied to your comment about the kebab casing for `config-server`, and I'm good with keeping the individual service managers.

lib/charms/hpc_libs/v0/slurm_ops.py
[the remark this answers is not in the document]
jedel1043: You're right, I misremembered the config name. Yeah, let's just map to `SLURMD_CONFIG_SERVER`.

NucciTheBoss:
On second thought, it would probably be better to implement the deb manager before bumping version. That way the lib is "complete" before building it into the Slurm charms.

PR #24, opened 2024-09-10
feat: implement `AptManager` for `slurm_ops`

jedel1043:
Implements an `AptManager` for `slurm_ops` that uses `apt` to install packages and `systemctl` to manage services.

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -518,37 +518,37 @@ class AptManager(SlurmOpsManager):
         slurm_wlm.import_key(
             textwrap.dedent(
                 """
-            -----BEGIN PGP PUBLIC KEY BLOCK-----
-            Comment: Hostname:
-            Version: Hockeypuck 2.2
+                -----BEGIN PGP PUBLIC KEY BLOCK-----
+                Comment: Hostname:
+                Version: Hockeypuck 2.2
 
-            xsFNBGTuZb8BEACtJ1CnZe6/hv84DceHv+a54y3Pqq0gqED0xhTKnbj/E2ByJpmT
-            NlDNkpeITwPAAN1e3824Me76Qn31RkogTMoPJ2o2XfG253RXd67MPxYhfKTJcnM3
-            CEkmeI4u2Lynh3O6RQ08nAFS2AGTeFVFH2GPNWrfOsGZW03Jas85TZ0k7LXVHiBs
-            W6qonbsFJhshvwC3SryG4XYT+z/+35x5fus4rPtMrrEOD65hij7EtQNaE8owuAju
-            Kcd0m2b+crMXNcllWFWmYMV0VjksQvYD7jwGrWeKs+EeHgU8ZuqaIP4pYHvoQjag
-            umqnH9Qsaq5NAXiuAIAGDIIV4RdAfQIR4opGaVgIFJdvoSwYe3oh2JlrLPBlyxyY
-            dayDifd3X8jxq6/oAuyH1h5K/QLs46jLSR8fUbG98SCHlRmvozTuWGk+e07ALtGe
-            sGv78ToHKwoM2buXaTTHMwYwu7Rx8LZ4bZPHdersN1VW/m9yn1n5hMzwbFKy2s6/
-            D4Q2ZBsqlN+5aW2q0IUmO+m0GhcdaDv8U7RVto1cWWPr50HhiCi7Yvei1qZiD9jq
-            57oYZVqTUNCTPxi6NeTOdEc+YqNynWNArx4PHh38LT0bqKtlZCGHNfoAJLPVYhbB
-            b2AHj9edYtHU9AAFSIy+HstET6P0UDxy02IeyE2yxoUBqdlXyv6FL44E+wARAQAB
-            zRxMYXVuY2hwYWQgUFBBIGZvciBVYnVudHUgSFBDwsGOBBMBCgA4FiEErocSHcPk
-            oLD4H/Aj9tDF1ca+s3sFAmTuZb8CGwMFCwkIBwIGFQoJCAsCBBYCAwECHgECF4AA
-            CgkQ9tDF1ca+s3sz3w//RNawsgydrutcbKf0yphDhzWS53wgfrs2KF1KgB0u/H+u
-            6Kn2C6jrVM0vuY4NKpbEPCduOj21pTCepL6PoCLv++tICOLVok5wY7Zn3WQFq0js
-            Iy1wO5t3kA1cTD/05v/qQVBGZ2j4DsJo33iMcQS5AjHvSr0nu7XSvDDEE3cQE55D
-            87vL7lgGjuTOikPh5FpCoS1gpemBfwm2Lbm4P8vGOA4/witRjGgfC1fv1idUnZLM
-            TbGrDlhVie8pX2kgB6yTYbJ3P3kpC1ZPpXSRWO/cQ8xoYpLBTXOOtqwZZUnxyzHh
-            gM+hv42vPTOnCo+apD97/VArsp59pDqEVoAtMTk72fdBqR+BB77g2hBkKESgQIEq
-            EiE1/TOISioMkE0AuUdaJ2ebyQXugSHHuBaqbEC47v8t5DVN5Qr9OriuzCuSDNFn
-            6SBHpahN9ZNi9w0A/Yh1+lFfpkVw2t04Q2LNuupqOpW+h3/62AeUqjUIAIrmfeML
-            IDRE2VdquYdIXKuhNvfpJYGdyvx/wAbiAeBWg0uPSepwTfTG59VPQmj0FtalkMnN
-            ya2212K5q68O5eXOfCnGeMvqIXxqzpdukxSZnLkgk40uFJnJVESd/CxHquqHPUDE
-            fy6i2AnB3kUI27D4HY2YSlXLSRbjiSxTfVwNCzDsIh7Czefsm6ITK2+cVWs0hNQ=
-            =cs1s
-            -----END PGP PUBLIC KEY BLOCK-----
-        """
+                xsFNBGTuZb8BEACtJ1CnZe6/hv84DceHv+a54y3Pqq0gqED0xhTKnbj/E2ByJpmT
+                NlDNkpeITwPAAN1e3824Me76Qn31RkogTMoPJ2o2XfG253RXd67MPxYhfKTJcnM3
+                CEkmeI4u2Lynh3O6RQ08nAFS2AGTeFVFH2GPNWrfOsGZW03Jas85TZ0k7LXVHiBs
+                W6qonbsFJhshvwC3SryG4XYT+z/+35x5fus4rPtMrrEOD65hij7EtQNaE8owuAju
+                Kcd0m2b+crMXNcllWFWmYMV0VjksQvYD7jwGrWeKs+EeHgU8ZuqaIP4pYHvoQjag
+                umqnH9Qsaq5NAXiuAIAGDIIV4RdAfQIR4opGaVgIFJdvoSwYe3oh2JlrLPBlyxyY
+                dayDifd3X8jxq6/oAuyH1h5K/QLs46jLSR8fUbG98SCHlRmvozTuWGk+e07ALtGe
+                sGv78ToHKwoM2buXaTTHMwYwu7Rx8LZ4bZPHdersN1VW/m9yn1n5hMzwbFKy2s6/
+                D4Q2ZBsqlN+5aW2q0IUmO+m0GhcdaDv8U7RVto1cWWPr50HhiCi7Yvei1qZiD9jq
+                57oYZVqTUNCTPxi6NeTOdEc+YqNynWNArx4PHh38LT0bqKtlZCGHNfoAJLPVYhbB
+                b2AHj9edYtHU9AAFSIy+HstET6P0UDxy02IeyE2yxoUBqdlXyv6FL44E+wARAQAB
+                zRxMYXVuY2hwYWQgUFBBIGZvciBVYnVudHUgSFBDwsGOBBMBCgA4FiEErocSHcPk
+                oLD4H/Aj9tDF1ca+s3sFAmTuZb8CGwMFCwkIBwIGFQoJCAsCBBYCAwECHgECF4AA
+                CgkQ9tDF1ca+s3sz3w//RNawsgydrutcbKf0yphDhzWS53wgfrs2KF1KgB0u/H+u
+                6Kn2C6jrVM0vuY4NKpbEPCduOj21pTCepL6PoCLv++tICOLVok5wY7Zn3WQFq0js
+                Iy1wO5t3kA1cTD/05v/qQVBGZ2j4DsJo33iMcQS5AjHvSr0nu7XSvDDEE3cQE55D
+                87vL7lgGjuTOikPh5FpCoS1gpemBfwm2Lbm4P8vGOA4/witRjGgfC1fv1idUnZLM
+                TbGrDlhVie8pX2kgB6yTYbJ3P3kpC1ZPpXSRWO/cQ8xoYpLBTXOOtqwZZUnxyzHh
+                gM+hv42vPTOnCo+apD97/VArsp59pDqEVoAtMTk72fdBqR+BB77g2hBkKESgQIEq
+                EiE1/TOISioMkE0AuUdaJ2ebyQXugSHHuBaqbEC47v8t5DVN5Qr9OriuzCuSDNFn
+                6SBHpahN9ZNi9w0A/Yh1+lFfpkVw2t04Q2LNuupqOpW+h3/62AeUqjUIAIrmfeML
+                IDRE2VdquYdIXKuhNvfpJYGdyvx/wAbiAeBWg0uPSepwTfTG59VPQmj0FtalkMnN
+                ya2212K5q68O5eXOfCnGeMvqIXxqzpdukxSZnLkgk40uFJnJVESd/CxHquqHPUDE
+                fy6i2AnB3kUI27D4HY2YSlXLSRbjiSxTfVwNCzDsIh7Czefsm6ITK2+cVWs0hNQ=
+                =cs1s
+                -----END PGP PUBLIC KEY BLOCK-----
+                """
             )
         )
 
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -562,37 +562,37 @@ class AptManager(SlurmOpsManager):
         experimental.import_key(
             textwrap.dedent(
                 """
-            -----BEGIN PGP PUBLIC KEY BLOCK-----
-            Comment: Hostname:
-            Version: Hockeypuck 2.2
+                -----BEGIN PGP PUBLIC KEY BLOCK-----
+                Comment: Hostname:
+                Version: Hockeypuck 2.2
 
-            xsFNBGTuZb8BEACtJ1CnZe6/hv84DceHv+a54y3Pqq0gqED0xhTKnbj/E2ByJpmT
-            NlDNkpeITwPAAN1e3824Me76Qn31RkogTMoPJ2o2XfG253RXd67MPxYhfKTJcnM3
-            CEkmeI4u2Lynh3O6RQ08nAFS2AGTeFVFH2GPNWrfOsGZW03Jas85TZ0k7LXVHiBs
-            W6qonbsFJhshvwC3SryG4XYT+z/+35x5fus4rPtMrrEOD65hij7EtQNaE8owuAju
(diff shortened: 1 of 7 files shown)


tests/integration/test_hpc_libs.yaml
```diff
@@ -38,11 +38,21 @@ acts:
           venv/bin/python3 -m pip install -r dev-requirements.txt
       - name: "Run `slurm_ops` integration tests"
         run: |
+          PYTHONPATH=./lib \
           venv/bin/python3 -m pytest -v \
             -s \
             --tb native \
             --log-cli-level=INFO \
             slurm_ops
+      - name: "Run `slurm_ops` integration tests (snap)"
```
NucciTheBoss, line 47: GitHub won't let me comment on line 39, but could we do the same thing with the act name we're doing here by having the act name for the deb version of the integration test be:

```yaml
- name: "Run `slurm_ops` integration tests (deb)"
```

That way folks can easily tell what the differences between the tests are in the integration test logs :smiley:

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -450,4 +464,192 @@ def _env_manager_for(self, type: ServiceType) -> _EnvManager:
...
+TbGrDlhVie8pX2kgB6yTYbJ3P3kpC1ZPpXSRWO/cQ8xoYpLBTXOOtqwZZUnxyzHh
+gM+hv42vPTOnCo+apD97/VArsp59pDqEVoAtMTk72fdBqR+BB77g2hBkKESgQIEq
+EiE1/TOISioMkE0AuUdaJ2ebyQXugSHHuBaqbEC47v8t5DVN5Qr9OriuzCuSDNFn
+6SBHpahN9ZNi9w0A/Yh1+lFfpkVw2t04Q2LNuupqOpW+h3/62AeUqjUIAIrmfeML
+IDRE2VdquYdIXKuhNvfpJYGdyvx/wAbiAeBWg0uPSepwTfTG59VPQmj0FtalkMnN
+ya2212K5q68O5eXOfCnGeMvqIXxqzpdukxSZnLkgk40uFJnJVESd/CxHquqHPUDE
+fy6i2AnB3kUI27D4HY2YSlXLSRbjiSxTfVwNCzDsIh7Czefsm6ITK2+cVWs0hNQ=
+=cs1s
+-----END PGP PUBLIC KEY BLOCK-----
+"""
```
NucciTheBoss, line 580: [nit]: Typically I like to have all constants declared at the beginning of the file after the import statements. It just flows better in my mind since I know where to refer to all constants within a particular module :sweat_smile:

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -450,4 +464,192 @@ def _env_manager_for(self, type: ServiceType) -> _EnvManager:
...
+        """
+        _call("systemctl", "enable", "--now", self._service.value)
+
+    def disable(self) -> None:
+        """Disable service."""
+        _call("systemctl", "disable", "--now", self._service.value)
+
+    def restart(self) -> None:
+        """Restart service."""
+        _call("systemctl", "reload-or-restart", self._service.value)
```
NucciTheBoss, line 493: [nit]: Could we potentially use a partial function here?

```python3
_systemctl = functools.partial(_call, "systemctl")
```

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -450,4 +464,192 @@ def _env_manager_for(self, type: ServiceType) -> _EnvManager:
...
+        proc = subprocess.run(
+            cmd,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.STDOUT,
+            text=True,
+            bufsize=1,
+            encoding="utf-8",
+        )
+        _logger.debug(f"command {cmd} exit code: {proc.returncode}. output:\n{proc.stdout}")
+        return proc.returncode == 0
```
NucciTheBoss, line 508: [nit]: I'm wondering if we should potentially just modify `_call` here to also return the exit code of a subprocess as well? Like `Tuple[str, int]`. That way we don't need a different way for calling `is-active`. Let me know your thoughts here.

_I know that this method is based off of the implementation from the systemd charm library_ :sweat_smile:

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -450,4 +464,192 @@ def _env_manager_for(self, type: ServiceType) -> _EnvManager:
...
+        override = Path("/etc/systemd/system/slurmd.service.d/10-slurmd-conf-server.conf")
+        override.parent.mkdir(exist_ok=True, parents=True)
+        override.write_text(
+            textwrap.dedent(
+                """
+                [Service]
+                ExecStart=/usr/bin/sh -c "/usr/sbin/slurmd -D -s $${SLURMD_CONFIG_SERVER:+--conf-server $$SLURMD_CONFIG_SERVER} $$SLURMD_OPTIONS"
+                """
+            )
+        )
```
NucciTheBoss, line 628: This looks good, but for the relevant `slurmd` bits, we should ensure that these methods don't conflict with what's been defined here for the slurmd service management utility in the slurmd charm:

https://github.com/charmed-hpc/slurm-charms/blob/246f1da25861766b0b42a020f3b40bff4bf20f32/charms/slurmd/src/utils/slurmd.py#L50-L96

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -450,4 +464,192 @@ def _env_manager_for(self, type: ServiceType) -> _EnvManager:
...
+6SBHpahN9ZNi9w0A/Yh1+lFfpkVw2t04Q2LNuupqOpW+h3/62AeUqjUIAIrmfeML
+IDRE2VdquYdIXKuhNvfpJYGdyvx/wAbiAeBWg0uPSepwTfTG59VPQmj0FtalkMnN
+ya2212K5q68O5eXOfCnGeMvqIXxqzpdukxSZnLkgk40uFJnJVESd/CxHquqHPUDE
+fy6i2AnB3kUI27D4HY2YSlXLSRbjiSxTfVwNCzDsIh7Czefsm6ITK2+cVWs0hNQ=
+=cs1s
+-----END PGP PUBLIC KEY BLOCK-----
+"""
+
+
+class AptManager(SlurmOpsManager):
```
NucciTheBoss, line 583: [nit]: Perhaps `DebManager` would be a better name here? Reason saying is that `AptManager` to me implies a generic apt interface for our hpc charms when what this class is really doing is managing the `slurm-wlm` (and friends) Debian package. Then our nomenclature is effectively `<package format>Manager`.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -64,17 +69,34 @@ def _on_install(self, _) -> None:
...
+    @property
+    def message(self) -> str:
+        """Return message passed as argument to exception."""
+        return self.args[0]
+
+
+try:
+    import charms.operator_libs_linux.v0.apt as apt
+except ImportError as e:
+    raise SlurmOpsError(f"could not import apt charm library. reason: {e}")
```
NucciTheBoss, line 98: Should we raise `SlurmOpsError` here if we fail to load the the apt charm library, or should we just "bubble up" the `ImportError` and have an error log message that the apt charm library isn't available and how to correct the issue? `ImportError` is still valid here as we're having "troubles" importing library dependencies.

```python3
try:
    import charms.operator_libs_linux.v0.apt as apt
except ImportError as e:
    _logger.error("down bad out here...")
   raise
```

Also, this way we don't have to define `SlurmOpsError` error before importing the apt charm library. _We'll just define the logger first instead_. Let me know what your thoughts are here.

NucciTheBoss (changes requested):
This looks great so far! :star_struck:  

Just a couple comments around the implementation of  `_MungectlManager` and `_AptManager`. Big thing I want to make sure that we're not doing is stepping into a minefield with overriding the systemd service file for slurmd both here and in the slurmd charm. Let me know if you have any questions.

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043: I don't think we should depend on `_logger` for error reporting, since people can just disable it globally.

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043: I also was considering this before, but I realized debs can also be managed by `dpkg`, and using the `<package manager>Manager` nomenclature better clarifies which specific tools we are using.

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043: I'll just move them inside the `install` function, since I think it's better to have them closer to its usage.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -382,11 +393,14 @@ def type(self) -> ServiceType:
         return self._service
 
 
-class _SnapMungeKeyManager(MungeKeyManager):
-    """Control the munge key using Snap."""
+class _MungectlManager(MungeKeyManager):
+    """Control the munge key using mungectl."""
+
+    def __init__(self, binpath: str) -> None:
+        self._binpath = binpath
```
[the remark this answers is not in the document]
jedel1043: Yeah, sounds like a cleaner solution

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043: If the problem is using `SlurmOpsError`, I can just use `ImportError` instead

lib/charms/hpc_libs/v0/slurm_ops.py
jedel1043: I think ensuring this should be the responsibility of the charms, and we should just give a sane default and document that the manager uses `/etc/default/slurmd` and `SLURMD_CONFIG_SERVER` to do that.

jedel1043: I'll still move the env vars redefinitions to different files, but the charm would need to ensure that the `SLURMD_CONFIG_SERVER` parameter is correctly set.

NucciTheBoss: Sounds good. We'll take note of this when we further integrate `slurm_ops` into the charms.

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: Since this is a nit, we can keep the class name as `AptManager`. With snap we have the benefit that the packager management tech and package format use the same name. When it comes to naming classes apt/deb management, it really just depends on your personal perspective :nerd_face:

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: Looking at the docs for [`ImportError`](https://docs.python.org/3/library/exceptions.html#ImportError), looks like we can only set the name of the module that we failed to import and the path to the module that's causing the problems. 

Given that we'd need to hack `ImportError` to include a message for how to install the `apt` charm library, let's just use `SlurmOpsError` here even if it feels funky to declare a custom exception before an import.

NucciTheBoss: _Ah wait_, it looks like you can pass a message to `ImportError`

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -448,6 +453,196 @@ def _env_manager_for(self, type: ServiceType) -> _EnvManager:
...
+            TbGrDlhVie8pX2kgB6yTYbJ3P3kpC1ZPpXSRWO/cQ8xoYpLBTXOOtqwZZUnxyzHh
+            gM+hv42vPTOnCo+apD97/VArsp59pDqEVoAtMTk72fdBqR+BB77g2hBkKESgQIEq
+            EiE1/TOISioMkE0AuUdaJ2ebyQXugSHHuBaqbEC47v8t5DVN5Qr9OriuzCuSDNFn
+            6SBHpahN9ZNi9w0A/Yh1+lFfpkVw2t04Q2LNuupqOpW+h3/62AeUqjUIAIrmfeML
+            IDRE2VdquYdIXKuhNvfpJYGdyvx/wAbiAeBWg0uPSepwTfTG59VPQmj0FtalkMnN
+            ya2212K5q68O5eXOfCnGeMvqIXxqzpdukxSZnLkgk40uFJnJVESd/CxHquqHPUDE
+            fy6i2AnB3kUI27D4HY2YSlXLSRbjiSxTfVwNCzDsIh7Czefsm6ITK2+cVWs0hNQ=
+            =cs1s
+            -----END PGP PUBLIC KEY BLOCK-----
+        """
```
NucciTheBoss, line 551: Looks like the formatting came out a bit funky here. Let's see if a suggestion can fix this:

```suggestion
                """
                -----BEGIN PGP PUBLIC KEY BLOCK-----
                Comment: Hostname:
                Version: Hockeypuck 2.2

                xsFNBGTuZb8BEACtJ1CnZe6/hv84DceHv+a54y3Pqq0gqED0xhTKnbj/E2ByJpmT
                NlDNkpeITwPAAN1e3824Me76Qn31RkogTMoPJ2o2XfG253RXd67MPxYhfKTJcnM3
                CEkmeI4u2Lynh3O6RQ08nAFS2AGTeFVFH2GPNWrfOsGZW03Jas85TZ0k7LXVHiBs
                W6qonbsFJhshvwC3SryG4XYT+z/+35x5fus4rPtMrrEOD65hij7EtQNaE8owuAju
                Kcd0m2b+crMXNcllWFWmYMV0VjksQvYD7jwGrWeKs+EeHgU8ZuqaIP4pYHvoQjag
                umqnH9Qsaq5NAXiuAIAGDIIV4RdAfQIR4opGaVgIFJdvoSwYe3oh2JlrLPBlyxyY
                dayDifd3X8jxq6/oAuyH1h5K/QLs46jLSR8fUbG98SCHlRmvozTuWGk+e07ALtGe
                sGv78ToHKwoM2buXaTTHMwYwu7Rx8LZ4bZPHdersN1VW/m9yn1n5hMzwbFKy2s6/
                D4Q2ZBsqlN+5aW2q0IUmO+m0GhcdaDv8U7RVto1cWWPr50HhiCi7Yvei1qZiD9jq
                57oYZVqTUNCTPxi6NeTOdEc+YqNynWNArx4PHh38LT0bqKtlZCGHNfoAJLPVYhbB
                b2AHj9edYtHU9AAFSIy+HstET6P0UDxy02IeyE2yxoUBqdlXyv6FL44E+wARAQAB
                zRxMYXVuY2hwYWQgUFBBIGZvciBVYnVudHUgSFBDwsGOBBMBCgA4FiEErocSHcPk
                oLD4H/Aj9tDF1ca+s3sFAmTuZb8CGwMFCwkIBwIGFQoJCAsCBBYCAwECHgECF4AA
                CgkQ9tDF1ca+s3sz3w//RNawsgydrutcbKf0yphDhzWS53wgfrs2KF1KgB0u/H+u
                6Kn2C6jrVM0vuY4NKpbEPCduOj21pTCepL6PoCLv++tICOLVok5wY7Zn3WQFq0js
                Iy1wO5t3kA1cTD/05v/qQVBGZ2j4DsJo33iMcQS5AjHvSr0nu7XSvDDEE3cQE55D
                87vL7lgGjuTOikPh5FpCoS1gpemBfwm2Lbm4P8vGOA4/witRjGgfC1fv1idUnZLM
                TbGrDlhVie8pX2kgB6yTYbJ3P3kpC1ZPpXSRWO/cQ8xoYpLBTXOOtqwZZUnxyzHh
                gM+hv42vPTOnCo+apD97/VArsp59pDqEVoAtMTk72fdBqR+BB77g2hBkKESgQIEq
                EiE1/TOISioMkE0AuUdaJ2ebyQXugSHHuBaqbEC47v8t5DVN5Qr9OriuzCuSDNFn
                6SBHpahN9ZNi9w0A/Yh1+lFfpkVw2t04Q2LNuupqOpW+h3/62AeUqjUIAIrmfeML
                IDRE2VdquYdIXKuhNvfpJYGdyvx/wAbiAeBWg0uPSepwTfTG59VPQmj0FtalkMnN
                ya2212K5q68O5eXOfCnGeMvqIXxqzpdukxSZnLkgk40uFJnJVESd/CxHquqHPUDE
                fy6i2AnB3kUI27D4HY2YSlXLSRbjiSxTfVwNCzDsIh7Czefsm6ITK2+cVWs0hNQ=
                =cs1s
                -----END PGP PUBLIC KEY BLOCK-----
                """
```

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -448,6 +453,196 @@ def _env_manager_for(self, type: ServiceType) -> _EnvManager:
...
```
NucciTheBoss, line 595: Looks like the formatting also got borked here as well. Suggestion to the rescue:

```suggestion
                """
                -----BEGIN PGP PUBLIC KEY BLOCK-----
                Comment: Hostname:
                Version: Hockeypuck 2.2

                xsFNBGTuZb8BEACtJ1CnZe6/hv84DceHv+a54y3Pqq0gqED0xhTKnbj/E2ByJpmT
                NlDNkpeITwPAAN1e3824Me76Qn31RkogTMoPJ2o2XfG253RXd67MPxYhfKTJcnM3
                CEkmeI4u2Lynh3O6RQ08nAFS2AGTeFVFH2GPNWrfOsGZW03Jas85TZ0k7LXVHiBs
                W6qonbsFJhshvwC3SryG4XYT+z/+35x5fus4rPtMrrEOD65hij7EtQNaE8owuAju
                Kcd0m2b+crMXNcllWFWmYMV0VjksQvYD7jwGrWeKs+EeHgU8ZuqaIP4pYHvoQjag
                umqnH9Qsaq5NAXiuAIAGDIIV4RdAfQIR4opGaVgIFJdvoSwYe3oh2JlrLPBlyxyY
                dayDifd3X8jxq6/oAuyH1h5K/QLs46jLSR8fUbG98SCHlRmvozTuWGk+e07ALtGe
                sGv78ToHKwoM2buXaTTHMwYwu7Rx8LZ4bZPHdersN1VW/m9yn1n5hMzwbFKy2s6/
                D4Q2ZBsqlN+5aW2q0IUmO+m0GhcdaDv8U7RVto1cWWPr50HhiCi7Yvei1qZiD9jq
                57oYZVqTUNCTPxi6NeTOdEc+YqNynWNArx4PHh38LT0bqKtlZCGHNfoAJLPVYhbB
                b2AHj9edYtHU9AAFSIy+HstET6P0UDxy02IeyE2yxoUBqdlXyv6FL44E+wARAQAB
                zRxMYXVuY2hwYWQgUFBBIGZvciBVYnVudHUgSFBDwsGOBBMBCgA4FiEErocSHcPk
                oLD4H/Aj9tDF1ca+s3sFAmTuZb8CGwMFCwkIBwIGFQoJCAsCBBYCAwECHgECF4AA
                CgkQ9tDF1ca+s3sz3w//RNawsgydrutcbKf0yphDhzWS53wgfrs2KF1KgB0u/H+u
                6Kn2C6jrVM0vuY4NKpbEPCduOj21pTCepL6PoCLv++tICOLVok5wY7Zn3WQFq0js
                Iy1wO5t3kA1cTD/05v/qQVBGZ2j4DsJo33iMcQS5AjHvSr0nu7XSvDDEE3cQE55D
                87vL7lgGjuTOikPh5FpCoS1gpemBfwm2Lbm4P8vGOA4/witRjGgfC1fv1idUnZLM
                TbGrDlhVie8pX2kgB6yTYbJ3P3kpC1ZPpXSRWO/cQ8xoYpLBTXOOtqwZZUnxyzHh
                gM+hv42vPTOnCo+apD97/VArsp59pDqEVoAtMTk72fdBqR+BB77g2hBkKESgQIEq
                EiE1/TOISioMkE0AuUdaJ2ebyQXugSHHuBaqbEC47v8t5DVN5Qr9OriuzCuSDNFn
                6SBHpahN9ZNi9w0A/Yh1+lFfpkVw2t04Q2LNuupqOpW+h3/62AeUqjUIAIrmfeML
                IDRE2VdquYdIXKuhNvfpJYGdyvx/wAbiAeBWg0uPSepwTfTG59VPQmj0FtalkMnN
                ya2212K5q68O5eXOfCnGeMvqIXxqzpdukxSZnLkgk40uFJnJVESd/CxHquqHPUDE
                fy6i2AnB3kUI27D4HY2YSlXLSRbjiSxTfVwNCzDsIh7Czefsm6ITK2+cVWs0hNQ=
                =cs1s
                -----END PGP PUBLIC KEY BLOCK-----
                """
```

NucciTheBoss (approved):
Great work on this!

I'm happy with how this library has come together. I made a couple "nit/foresight"-level comments, but these are things that can be addressed in the future as part of gradual improvements to the Slurm charms

There are some formatting issues with the PPA keys, so I'm just going to fix them quickly with suggestions, and then open a PR to bump the version of `slurm_ops` for Charmhub.

PR #26, opened 2024-09-16
fix: slurm-prometheus-exporter -> prometheus-slurm-exporter

### LINKED ISSUE #25: [Bug]: Incorrect service name for Slurm prometheus exporter
The current name we have for the Slurm Prometheus exporter in `slurm_ops` goes against Debian's convention for Prometheus exporters - `slurm-prometheus-exporter` rather than the standard `prometheus-slurm-exporter` - and the service name provided by our Debian package is `prometheus-slurm-exporter`

Quick fix. We just need to update the `ServiceType` enum to use the correct service name.

NucciTheBoss:
Small PR to change exporter service name from `slurm-prometheus-exporter` to `prometheus-slurm-exporter` so that the service name is standard with other exporters packaged for Debian.

Closes #25

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -156,7 +156,7 @@ class ServiceType(Enum):
     """Type of Slurm service to manage."""
 
     MUNGED = "munged"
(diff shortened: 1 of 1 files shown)


jedel1043 (approved):
LGTM!

NucciTheBoss:
@jedel1043 go back to Independence Day fam :mexico:

PR #32, opened 2024-09-18
fix(slurm_ops): ensure that all slurm daemons are installed

### LINKED ISSUE #28: [Bug]: `AptManager` only installs slurmctld and slurmd, but not slurmdbd and slurmrestd
On my crusade to implement `slurm_ops` within the Slurm charms, I realized that we're only installing `slurmctld` and `slurmd` by pulling the `slurm-wlm` package. You can verify this by installing the `slurm-wlm` package and then run the following commands:

```shell
apt update && apt install slurm-wlm
systemctl status slurmdbd
systemctl status slurmrestd
```

Not difficult to fix, we just need to expand the list of packages we're installing to include slurmdbd and slurmrestd:

```python3
apt.add_package(["slurmctld", "slurmd", "slurmdbd", "slurmrestd", "slurm-client", "mungectl", "prometheus-slurm-exporter"])
```

NucciTheBoss:
This PR ensures that all the Slurm daemons are installed on the host with the necessary client commands. This mimics the behavior of the Slurm snap where all the daemons are installed but are controlled through orchestration with the charms. This also helps that our version of Slurm is consistent across all machines since Slurm doesn't guarantee backwards compatibility between daemon versions.

Fixes #28

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -461,7 +461,15 @@ class _AptManager(_OpsManager):
         repositories.add(experimental)
 
         apt.update()
-        for package in ["slurm-wlm", "mungectl", "prometheus-slurm-exporter"]:
+        for package in [
+            "slurmctld",
+            "slurmd",
+            "slurmdbd",
+            "slurmrestd",
+            "slurm-client",
+            "mungectl",
+            "prometheus-slurm-exporter",
+        ]:
             try:
                 apt.add_package(package)
             except apt.PackageNotFoundError as e:


lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -461,7 +461,15 @@ def install(self) -> None:
...
-        for package in ["slurm-wlm", "mungectl", "prometheus-slurm-exporter"]:
+        for package in [
+            "slurmctld",
+            "slurmd",
+            "slurmdbd",
+            "slurmrestd",
+            "slurm-client",
+            "mungectl",
+            "prometheus-slurm-exporter",
+        ]:
```
jedel1043, line 472: After merging the `AptManager` I realized that `apt.add_package` already puts on its error the list of packages that fail to get installed, so I think we can move this array to be inside `apt.add_package`, maybe using a generic message for the raised `SlurmOpsError` instead of printing the package.

NucciTheBoss:
Hmm... forgot to catch that slurm-wlm is the package being used to get the version of Slurm installed on the host. Can switch to slurm-client instead as that package is installed on all the nodes.

jamesbeedy:
We should omit the slurm-client package from slurmdbd and slurmrestd. Can we use the specific slurm package that is installed for each respective daemon?

jamesbeedy:
Is there a reason to add all of the slurm daemons to every node?

NucciTheBoss:
Hey hey @jamesbeedy - thanks for chiming in here!

> We should omit the slurm-client package from slurmdbd and slurmrestd. Can we use the specific slurm package that is installed for each respective daemon?

Omitting the `slurm-client` package is possible; I mostly added it because (a) we can use it in place of the `slurm-wlm` package to grab the version of Slurm installed on the units and (b) being able to talk to Slurm from anywhere in your cluster is nice. We can play around with it however since slurm-client is brought in as a dependency on the units that need the client commands to be available.

> Is there a reason to add all of the slurm daemons to every node?

As for installing the Slurm daemons on each node with `_AptManager`, this is to emulate the behavior of the Slurm snap. All the daemons are provided by the Slurm snap, and through the charms we can control which daemon is started up. Makes it easier for ensuring a consistent Slurm version/API across the entire cluster. We take the "kitchen sink" approach here since it makes abstracting the differences between how the formats are handled easier to work with. We can play around with it however. Originally we were just using the `slurm-wlm` to install the daemons.

Eventually where we want to get is that we can provide a base Slurm implementation that anyone can use, but then also support side-loading your own version of Slurm for those that require custom extensions.

NucciTheBoss:
@jedel1043 @jamesbeedy commit https://github.com/charmed-hpc/hpc-libs/pull/32/commits/1fe995bd2d6b00ba51738245f7accdb564aa0206 enables `_AptManager` to manage a specific Slurm daemon rather than the "kitchen sink". Let me know what you think!

jedel1043 (approved):
LGTM!

NucciTheBoss:
Thanks @jedel1043 @jamesbeedy - will hold off on vbump after I add the necessary bits for jwt and cgroup editing.

PR #34, opened 2024-10-01
feat(slurm_ops): implement initial jwt key manager

### LINKED ISSUE #10: [Enhancement]: Add JWT key handling support
We should add JWT key handling support to `SlurmManagerBase` similar to how we handle munge keys. I went back and forth on this a bit as not all the Slurm daemons support using JWT keys - munge the OG - but I don't think it would hurt us in the long run as `slurmctld`, `slurmdbd`, and `slurmrestd` support using JWT keys.

We can enable JWT key support using composition, but without inheriting from the `Service` class:

```python3
class SlurmManagerBase(Service):
    def __init__(self, service: ServiceType) -> None:
        self.munge = MungeManager()
        self.jwt = JWTManager()
        ...
```

I'm thinking that we add a `jwtctl` tool similar `mungectl`, where we can generate, set, and get keys from the command line. That way we directly manage the key in the snap package, and folks can still benefit from the tools that we are creating even if they're not using the charm or snap package.

Let me know you thoughts @jedel1043 :eyes:

NucciTheBoss:
This PR introduces initial support for managing JWT keys within `slurm_ops`. It's similar to the `_MungeKeyManager` class, but with the exception that it uses the [`cryptography`](https://cryptography.io/en/latest/) module to generate the key rather than a dedicated tool. Ideally we'll eventually use a dedicated tool - something like `jwtctl` - but that's for next cycle.

Also, this PR makes the small change of generating a `StateSaveLocation` directory for Slurm. The Debian package doesn't do this, so after install the `_AptManager` will create the `var/lib/slurm/slurm.state` directory that can be used for stashing the key and then exporting across the cluster. There doesn't seem to be an agreed upon standard, so I just chose the dirname `slurm.state` since it's clear that this is the directory where the slurm daemons will save their state. The snap uses `checkpoint` as the name, but we can easily change this.

Closes #10

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -73,6 +73,8 @@ from typing import Any, Optional, Union
 import distro
 import dotenv
 import yaml
+from cryptography.hazmat.primitives import serialization
+from cryptography.hazmat.primitives.asymmetric import rsa
 from slurmutils.editors import cgroupconfig, slurmconfig, slurmdbdconfig
 from slurmutils.models import CgroupConfig, SlurmConfig, SlurmdbdConfig
 
@@ -96,7 +98,13 @@ LIBAPI = 0
 LIBPATCH = 7
 
 # Charm library dependencies to fetch during `charmcraft pack`.
-PYDEPS = ["pyyaml>=6.0.2", "python-dotenv~=1.0.1", "slurmutils~=0.7.0", "distro~=1.9.0"]
+PYDEPS = [
+    "cryptography~=43.0.1",
+    "pyyaml>=6.0.2",
+    "python-dotenv~=1.0.1",
+    "slurmutils~=0.7.0",
+    "distro~=1.9.0",
+]
 
 _logger = logging.getLogger(__name__)
 
@@ -549,6 +557,9 @@ class _AptManager(_OpsManager):
             raise SlurmOpsError(f"failed to install {self._service_name}. reason: {e}")
 
         self._env_file.touch(exist_ok=True)
+        # Debian package postinst hook does not create a `StateSaveLocation` directory
+        # so we make one here that is only r/w by owner.
+        Path("/var/lib/slurm/slurm.state").mkdir(mode=0o600, exist_ok=True)
 
         if self._service_name == "slurmd":
             override = Path("/etc/systemd/system/slurmd.service.d/10-slurmd-conf-server.conf")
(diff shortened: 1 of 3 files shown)


NucciTheBoss:
Works on my machine :skull:

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -584,6 +595,42 @@ def _env_manager_for(self, type: _ServiceType) -> _EnvManager:
...
+        JWT key used by `slurmctld` and `slurmrestd`. This way we won't need to
+        pass the `snap` bool as an argument to `__init__`
+    """
+
+    def __init__(self, snap: bool = False):
+        self._keyfile_path = (
+            Path("/var/snap/slurm/common/var/lib/slurm/slurm.state/jwt_hs256.key")
+            if snap
+            else Path("/var/lib/slurm/slurm.state/jwt_hs256.key")
+        )
```
jedel1043, line 612: This could be a bit less dependent on the snap option by adding a `state_save_location` method to `_OpsManager`, then making our managers return their preferred path for that.

jedel1043 (commented):
Nice work! Some small comments

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -584,6 +595,42 @@ def _env_manager_for(self, type: _ServiceType) -> _EnvManager:
         return _EnvManager(file=self._env_file, prefix=type.value)
 
 
+class _JWTKeyManager:
+    """Control the JWT key used by Slurm.
+
+    Todo:
+        Use `jwtctl` to provide backend for generating, setting, and getting
+        JWT key used by `slurmctld` and `slurmrestd`. This way we won't need to
+        pass the `snap` bool as an argument to `__init__`
```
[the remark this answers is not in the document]
NucciTheBoss: Private classes like `_JWTKeyManager` are not rendered as part of the public documentation for [`slurm_ops`](https://charmhub.io/hpc-libs/libraries/slurm_ops) on Charmhub; only the docstrings for public classes/functions are shown for charm libraries.

However, I think your comment here indicates that we need an agreeable standard/practice for having TODO comments in our code. As our portfolio grows, so will the times where we'll need to implement short-term solutions that "work" but aren't how we want it to be in the future. _I'm thinking about our discussion from earlier today about how we need to spec out the best-practices we follow._

Since we already follow the Google docstring format, perhaps we also follow the TODO styleguide as well? https://google.github.io/styleguide/pyguide.html#312-todo-comments. Let me know what you think @jedel1043 :nerd_face:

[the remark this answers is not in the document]
jedel1043: Sounds good to me!

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: Hmm... I think what we really need here is a property such as `var_lib_path` that returns which `/var/lib` path we should be using. Other Slurm configuration options will need to know what the correct `/var/lib` path is - `StateSaveLocation` is shared by all the Slurm daemons - so it might make sense here to expose common paths and let the managers perform the relevant manipulations that they need to.

jedel1043: Yep, `var_lib_path` sounds like a nicer solution

NucciTheBoss:
@jedel1043 R4R

I created issues to help track the work mentioned in the TODO comments such as switch to the Slurm snap stable channel and creating `jwtctl` to help with managing the jwt signing key used by Slurm.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -680,7 +693,7 @@ def __init__(self, service: _ServiceType, snap: bool = False) -> None:
         self._ops_manager = _SnapManager() if snap else _AptManager(service)
         self.service = self._ops_manager.service_manager_for(service)
         self.munge = _MungeManager(self._ops_manager)
-        self.jwt = _JWTKeyManager(snap)
+        self.jwt = _JWTKeyManager(self._ops_manager.var_lib_path / "slurm.data/jwt_hs256.key")
```
jedel1043, line 696: nit: we could move this to `_JWTKeyManager` and only pass the manager, to have everything about jwt encapsulated in a single class.

jedel1043 (approved):
Looks really nice! I have a small nitpick that doesn't block merging this as it is.

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: Sure thing, not a huge issue for me. This also reduces the chances that we need to modify the constructor for `_JWTKeyManager` in the future.

NucciTheBoss:
@jedel1043 one more review to make sure we like the `_JWTKeyManager` implementation with `_OpsManager` :pleading_face::point_right::point_left:

jedel1043 (approved):
Looks perfect!

PR #38, opened 2024-10-01
fix: format -> fmt

NucciTheBoss:
All of the Charmed HPC tox instances use `fmt`, not `format`. Switching back to `fmt` to that devx is consistent across Charmed HPC repos.

Misc. changes:

- Add periods to the end of each environment description.

Also, this PR highlights that we need to put together a spec for what HPC team tox files should look like. Should make some time at either leading up to roadmap or during the engineering sprint when we have the team together in-person.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -4,7 +4,7 @@
 [tox]
 no_package = True
 skip_missing_interpreters = True
-env_list = format, lint, static, unit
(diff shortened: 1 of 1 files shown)


jedel1043 (approved):
LGTM!

PR #44, opened 2024-10-10
feat(slurm_ops): loosen environment variable rules and configure openfile limits

### LINKED ISSUE #41: [Enhancement]: openfiles not configured for hpc use case
Ulimit settings of the operating system (`/etc/security/limits.conf`) and for processes `slurmd` and `slurmctld` restrict `slurmscriptd` processes' openfiles to 1048.

```bash
$ srun -pslurmd --pty  bash
ubuntu@juju-1cc933-0:~$ ulimit -n
1048
```

### LINKED ISSUE #43: [Enhancement]: Loosen up `_EnvManager` environment variable rules
The `_EnvManager` class currently prefixes environment variables with the service name in SCREAMING CASE. This is great for for when all environment variables are managed within a single file - _one .env to rule them all_ - as you can easily tell which environment variable is used by each service; however, this doesn't work in all cases.

The specific case where this implementation breaks down is with `slurmdbd`. Why you ask? Because `slurmdbd` doesn't support configuring a database backend where you communicate with MySQL via a unix socket rather than a TCP endpoint. You must set the environment variable _MYSQL_UNIX_PORT_ for `slurmdbd` to use the unix socket, otherwise it will just fail to connect to the database. If we try to use `_EnvManager` to set the socket, it will prefix the env var as `SLURMDBD_MYSQL_UNIX_PORT` which won't be recognized by `slurmdbd`.

Long term solution is to patch Slurm to support configuring this option - _doesn't seem that difficult but takes time & commitment_ - so for now we should just loosen up the formatting rules a bit in `_EnvManager`.

NucciTheBoss:
This PR adds support for setting the `MYSQL_UNIX_PORT` environment variable needed by slurmdbd to connect to MySQL via a unix socket, and it adds configs for expanding the allowed amount open files at once.

This PR also improves the testing coverage of `_AptManager` by mocking several of the installation steps such as configuring apt to use the Ubuntu HPC PPA, writing security configurations, overriding service files, etc. It refactors the install function to (a) lower the cyclomatic complexity of `install` - less if/else in a single function - and (b) very easy to unit test.

### Misc.

* Configure black's target version to be py310 so that we can use match/case. [Python 3.8 is also EOL](https://endoflife.date/python), and we're not targeting any bases that use Python 3.8 by default.
* Simplify unit tests by using only one `TestCase` class. `pyfakefs`'s version of `TestCase` inherits from `unittest.TestCase`, so just don't call `self.setupPyfakeFS()` and it's pretty much the same :sweat: 

## Related issues

* Closes #43 
* Closes #41

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -482,7 +482,7 @@ class _AptManager(_OpsManager):
         self._install_service()
         # Debian package postinst hook does not create a `StateSaveLocation` directory
         # so we make one here that is only r/w by owner.
-        _logger.debug("creating slurm statesavelocation directory")
+        _logger.debug("creating slurm `StateSaveLocation` directory")
         Path("/var/lib/slurm/slurm.state").mkdir(mode=0o600, exist_ok=True)
         self._apply_overrides()
 
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -478,6 +478,47 @@ class _AptManager(_OpsManager):
 
     def install(self) -> None:
         """Install Slurm using the `slurm` snap."""
+        self._init_ubuntu_hpc_ppa()
+        self._install_service()
+        # Debian package postinst hook does not create a `StateSaveLocation` directory
+        # so we make one here that is only r/w by owner.
+        _logger.debug("creating slurm statesavelocation directory")
+        Path("/var/lib/slurm/slurm.state").mkdir(mode=0o600, exist_ok=True)
+        self._apply_overrides()
+
+    def version(self) -> str:
+        """Get the current version of Slurm installed on the system."""
+        try:
+            return apt.DebianPackage.from_installed_package(self._service_name).version.number
+        except apt.PackageNotFoundError as e:
+            raise SlurmOpsError(f"unable to retrieve {self._service_name} version. reason: {e}")
+
+    @property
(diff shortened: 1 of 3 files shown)


NucciTheBoss:
@jedel1043 R4R now that we're not getting ganked by Python

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -483,6 +478,47 @@ def __init__(self, service: _ServiceType) -> None:
 
     def install(self) -> None:
         """Install Slurm using the `slurm` snap."""
+        self._init_ubuntu_hpc_ppa()
+        self._install_service()
+        # Debian package postinst hook does not create a `StateSaveLocation` directory
+        # so we make one here that is only r/w by owner.
+        _logger.debug("creating slurm statesavelocation directory")
```
jedel1043, line 485:
```suggestion
        _logger.debug("creating slurm StateSaveLocation directory")
```

Nit: we should use the "proper" name in logs.

jedel1043 (approved):
Really impressive work! I have a very tiny nitpick that shouldn't block merging this :)

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: Tbh I was on the fence about keeping this entirely lowercase, but since you call it out, I'm fine with just quickly patching this and then merging the PR :smile:

NucciTheBoss:
_Oops_ - latest commit message got messed up. I'll fix in a vbump PR tomorrow.

PR #45, opened 2024-10-10
ci: fix CI pipeline and add TiCS analysis action

jedel1043:
- Fixes the CI pipeline, since the recent upgrade to 24.04 broke it.
- Adds a TiCS action that runs once every day.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -9,7 +9,7 @@ on:
 jobs:
   inclusive-naming-check:
     name: Inclusive naming check
-    runs-on: ubuntu-latest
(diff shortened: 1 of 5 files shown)


NucciTheBoss (approved):
LGTM!

`pipx` works for now as it's maintained by pypa. One thing we'll want to possibly look at is using the tox package from the Ubuntu Archive. We're currently not pinning the version, so if there's ever a tox 5 (or something gets deprecated), we'll need to keep an eye out for that.

Either way, this is part of a larger discussion about GitHub (like issue labels + repo access rules), so maybe we'll create a template repo where we can pull files from.

PR #55, opened 2024-11-20
feat(slurm_ops)!: implement SackdManager for `sackd` service

NucciTheBoss:
> This PR is currently a draft because the apt charm library does not work on Noble based machines. Related bug report: https://github.com/canonical/operator-libs-linux/issues/135.
>
> Will do the version bump after apt charm library issues are fixed and the PR has been approved.

This PR introduces the `SackdManager` class which can be used in a downstream sackd-operator to manage the `sackd` service on machines.

BREAKING CHANGES: This PR drops the PPA `ubuntu-hpc/slurm-wlm-23.02` as it doesn't provide a `sackd` package, and the Slurm version that we can pull from Universe on Noble is newer than what is provided by the PPA. The `ubuntu-hpc/slurm-wlm-23.02` PPA also doesn't provide any packages that work on Noble.

Dropping this PPA means that you will now get an older version of Slurm on Jammy machines, but that isn't a huge issue as we're planning to push the Slurm charms to Noble anyway.

### Misc.

Also this PR makes it such that `prometheus-slurm-exporter` is only installed on `slurmctld` machines since the slurmctld operator is the only Slurm charm that currently integrates with COS via the `cos-agent` relation. Helps minimize our install size just a little bit.

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -635,10 +635,10 @@ class _AptManager(_OpsManager):
         Raises:
             SlurmOpsError: Raised if `apt` fails to install the required Slurm packages.
         """
-        packages = [self._service_name, "munge", "mungectl", "prometheus-slurm-exporter"]
+        packages = [self._service_name, "munge", "mungectl"]
         match self._service_name:
             case "slurmctld":
-                packages.extend(["libpmix-dev", "mailutils"])
+                packages.extend(["libpmix-dev", "mailutils", "prometheus-slurm-exporter"])
             case "slurmd":
                 packages.extend(["libpmix-dev", "openmpi-bin"])
             case "slurmrestd":
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -51,6 +51,7 @@ class ApplicationCharm(CharmBase):
 """
 
 __all__ = [
+    "SackdManager",
     "SlurmOpsError",
     "SlurmctldManager",
     "SlurmdManager",
@@ -179,6 +180,7 @@ class _ServiceType(Enum):
 
     MUNGE = "munge"
     PROMETHEUS_EXPORTER = "prometheus-slurm-exporter"
+    SACKD = "sackd"
     SLURMD = "slurmd"
     SLURMCTLD = "slurmctld"
     SLURMDBD = "slurmdbd"
@@ -425,12 +427,12 @@ class _SnapManager(_OpsManager):
     def install(self) -> None:
         """Install Slurm using the `slurm` snap."""
         # TODO: https://github.com/charmed-hpc/hpc-libs/issues/35 -
-        #  Pin Slurm snap to stable channel.
+        #   Pin Slurm snap to stable channel.
         _snap("install", "slurm", "--channel", "latest/candidate", "--classic")
         # TODO: https://github.com/charmed-hpc/slurm-snap/issues/49 -
-        #  Request automatic alias for the Slurm snap so we don't need to do it here.
-        #  We will possibly need to account for a third-party Slurm snap installation
-        #  where aliasing is not automatically performed.
+        #   Request automatic alias for the Slurm snap so we don't need to do it here.
+        #   We will possibly need to account for a third-party Slurm snap installation
+        #   where aliasing is not automatically performed.
         _snap("alias", "slurm.mungectl", "mungectl")
 
     def version(self) -> str:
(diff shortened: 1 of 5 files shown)


lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -908,6 +890,29 @@ def scontrol(*args) -> str:
         return _call("scontrol", *args).stdout
 
 
+class SackdManager(_SlurmManagerBase):
+    """Manager for the Sackd service."""
```
dsloanm, line 894: Does SACKD_CONFIG_SERVER need documented here as SLURMD_CONFIG_SERVER is for SlurmdManager?

NucciTheBoss: It should be documented! I'm going to clean up the docstrings while the work on enabling the `apt` charm library to handle deb822 formatted sources.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -635,10 +593,12 @@ def _install_service(self) -> None:
         Raises:
             SlurmOpsError: Raised if `apt` fails to install the required Slurm packages.
         """
-        packages = [self._service_name, "munge", "mungectl", "prometheus-slurm-exporter"]
+        packages = [self._service_name, "munge", "mungectl"]
         match self._service_name:
+            case "sackd":
+                packages.extend(["slurm-client"])
```
dsloanm, line 599: Also "libpmix-dev" and "openmpi-bin" to let users build MPI codes and to keep the environment consistent with the compute nodes?

jamesbeedy: Hey hey, libpmix-dev is a runtime dependency for slurmd and slurmctld. Openmpi-bin is added to compute nodes so they can run mpi workloads. I don't think either of these packages belong on the login node.

dsloanm: Hello! Definitely agreed MPI workloads shouldn't run on the login node. My thought process for having OpenMPI was to allow users to compile MPI binaries for use on the compute nodes since that's commonly done on login nodes (though it'd need "libopenmpi-dev" since "openmpi-bin" provides only `mpicc`, etc. compiler wrappers and not `mpi.h`).

Maybe worth revisiting in a future PR. Happy to keep this one focused on `SackdManager` and stick with "slurm-client".

NucciTheBoss: For now we should have `slurm-client` be the only additional package on the login node.

When we put together some howtos for common HPC use cases such as how to submit jobs, that would be a great time to identify other packages we should be shipping on the nodes. Our software stack story is also still loosely defined, so it might not even be the `sackd` charm itself providing the packages folks will load and use.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -674,6 +634,28 @@ def _create_state_save_location(self) -> None:
     def _apply_overrides(self) -> None:
         """Override defaults supplied provided by Slurm Debian packages."""
         match self._service_name:
+            case "sackd":
+                _logger.debug("overriding default sackd service configuration")
+                config_override = Path(
+                    "/etc/systemd/system/sackd.service.d/10-sackd-config-server.conf"
+                )
+                config_override.mkdir(parents=True, exist_ok=True)
```
dsloanm, line 642: Spotted in testing: this should be `config_override.parent.mkdir`.

lib/charms/hpc_libs/v0/slurm_ops.py
```diff
@@ -674,6 +634,28 @@ def _create_state_save_location(self) -> None:
...
+                config_override.write_text(
+                    textwrap.dedent(
+                        """
+                        [Service]
+                        ExecStart=
+                        ExecStart=/usr/bin/sh -c "/usr/sbin/sackd --systemd $${SACKD_CONFIG_SERVER:+--conf-server $$SACKD_CONFIG_SERVER} $$SACKD_OPTIONS"
+                        """
```
dsloanm, line 649: Could we make this:

```
"""
[Service]
ExecStart=
ExecStart=/usr/bin/sh -c "/usr/sbin/sackd $${SACKD_CONFIG_SERVER:+--conf-server $$SACKD_CONFIG_SERVER} $$SACKD_OPTIONS"
Type=forking
"""
```

At the moment, sackd.service fails to start with:

```
systemd[1]: sackd.service: Got notification message from PID 7809, but reception only permitted for main PID 7808
systemd[1]: sackd.service: start operation timed out. Terminating.
```

Adding `Type=forking` gets you to `sh[7101]: sackd: fatal: Missing NOTIFY_SOCKET` then removing `--systemd` allows the service to start.

Alternatively, setting `SACKD_OPTIONS="--conf-server <hostname>:6817"` and having no config override could be an idea (e.g. through `config_server`?)

NucciTheBoss: Hmm... not sure we should use Type=forking here. Mostly because it's discouraged by systemd: https://www.freedesktop.org/software/systemd/man/latest/systemd.service.html#Options

I think we can do some magic in the sackd environment file to get around needing to start sackd up in a child process that needs to be monitored.

dsloanm: Agreed with the environment file. Getting the config into `SACKD_OPTIONS` in `/etc/default/sackd` means we wouldn't need any override and could remove the `case "sackd":` branch of `_apply_overrides()` entirely.

Similarly, the `slurmd` case could have its options moved to `SLURMD_OPTIONS` in `/etc/default/sackd`, and [its use of `Type=forking` reassessed](https://github.com/charmed-hpc/slurm-charms/blob/main/charms/slurmd/src/utils/service.py#L55), I say cautiously, not wanting to open the `service.py` can of worms...

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: Fixed in latest batch of commits. Took some time to improve the docstrings because some were out of date and providing inaccurate information.

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: Only installing `slurm-client` with `sackd` for now. Will see what else should be added to the nodes once we dive more into the "providing dependencies for user workloads" story.

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: Fixed in latest batch of commits.

lib/charms/hpc_libs/v0/slurm_ops.py
NucciTheBoss: I got it such that we don't need to use `Type=forking` for the daemon. Rather than perform the variable substitution directly in the service file, we instead do it the _/etc/default/sackd_ file used by the systemd service file.

```shell
SACKD_CONFIG_SERVER=''
SACKD_OPTIONS="${SACKD_CONFIG_OPTIONS:+--conf-server $SACKD_CONFIG_SERVER}"
``` 

Key thing to note is that `SACKD_CONFIG_SERVER` will always need to come before `SACKD_OPTIONS`, otherwise the variable substitution will not be performed correctly when sourcing in the environment file.

dsloanm: Hmm, this isn't working in my testing. The `sackd` service fails to start when _/etc/default/sackd_ is:

```
SACKD_CONFIG_SERVER='juju-46b4c3-0:6817'
SACKD_OPTIONS="${SACKD_CONFIG_SERVER:+--conf-server $SACKD_CONFIG_SERVER}"
```

but starts if I manually edit it to:

```
SACKD_OPTIONS="--conf-server juju-46b4c3-0:6817"
```

Details below:

```
ubuntu@juju-46b4c3-1:~$ sudo systemctl restart sackd
Job for sackd.service failed because the control process exited with error code.
See "systemctl status sackd.service" and "journalctl -xeu sackd.service" for details.
ubuntu@juju-46b4c3-1:~$ sudo systemctl status sackd
× sackd.service - Slurm auth and cred kiosk daemon
     Loaded: loaded (/usr/lib/systemd/system/sackd.service; enabled; preset: enabled)
     Active: failed (Result: exit-code) since Fri 2024-12-06 18:03:07 UTC; 1s ago
       Docs: man:sackd(8)
    Process: 5262 ExecStart=/usr/sbin/sackd --systemd $SACKD_OPTIONS (code=exited, status=1/FAILURE)
   Main PID: 5262 (code=exited, status=1/FAILURE)
        CPU: 6ms

Dec 06 18:03:07 juju-46b4c3-1 systemd[1]: Starting sackd.service - Slurm auth and cred kiosk daemon...
Dec 06 18:03:07 juju-46b4c3-1 sackd[5262]: sackd: error: resolve_ctls_from_dns_srv: res_nsearch error: Unknown host
Dec 06 18:03:07 juju-46b4c3-1 sackd[5262]: sackd: error: fetch_config: DNS SRV lookup failed
Dec 06 18:03:07 juju-46b4c3-1 sackd[5262]: sackd: fatal: _establish_config_source: failed to load configs
Dec 06 18:03:07 juju-46b4c3-1 systemd[1]: sackd.service: Main process exited, code=exited, status=1/FAILURE
Dec 06 18:03:07 juju-46b4c3-1 systemd[1]: sackd.service: Failed with result 'exit-code'.
Dec 06 18:03:07 juju-46b4c3-1 systemd[1]: Failed to start sackd.service - Slurm auth and cred kiosk daemon.
ubuntu@juju-46b4c3-1:~$ sudo cat /etc/default/sackd 
SACKD_CONFIG_SERVER='juju-46b4c3-0:6817'
SACKD_OPTIONS="${SACKD_CONFIG_SERVER:+--conf-server $SACKD_CONFIG_SERVER}"
ubuntu@juju-46b4c3-1:~$ sudo nano /etc/default/sackd 
ubuntu@juju-46b4c3-1:~$ sudo cat /etc/default/sackd 
SACKD_OPTIONS="--conf-server juju-46b4c3-0:6817"
ubuntu@juju-46b4c3-1:~$ sudo systemctl restart sackd
ubuntu@juju-46b4c3-1:~$ sudo systemctl status sackd
● sackd.service - Slurm auth and cred kiosk daemon
     Loaded: loaded (/usr/lib/systemd/system/sackd.service; enabled; preset: enabled)
     Active: active (running) since Fri 2024-12-06 18:04:28 UTC; 3s ago
       Docs: man:sackd(8)
   Main PID: 5323 (sackd)
      Tasks: 11
     Memory: 1.6M (peak: 2.0M)
        CPU: 9ms
     CGroup: /system.slice/sackd.service
             └─5323 /usr/sbin/sackd --systemd --conf-server juju-46b4c3-0:6817

Dec 06 18:04:28 juju-46b4c3-1 systemd[1]: Starting sackd.service - Slurm auth and cred kiosk daemon...
Dec 06 18:04:28 juju-46b4c3-1 sackd[5323]: sackd: running
Dec 06 18:04:28 juju-46b4c3-1 systemd[1]: Started sackd.service - Slurm auth and cred kiosk daemon.
ubuntu@juju-46b4c3-1:~$
```

NucciTheBoss: In the new commit https://github.com/charmed-hpc/hpc-libs/pull/55/commits/1b967905297d813a2dd0731bd2b153054b13286d, @jedel1043 and I decided that we should just mandate that the configuration server is provided rather than try to be clever with shell variable substitution. This way we can ensure that `sackd` is started the way we want it started, and we don't need to make `sackd` `Type=forking`.

Here's new implementation working in a my development playground.

```
● sackd.service - Slurm auth and cred kiosk daemon
     Loaded: loaded (/usr/lib/systemd/system/sackd.service; enabled; preset: enabled)
    Drop-In: /etc/systemd/system/sackd.service.d
             └─10-sackd-config-server.conf
     Active: active (running) since Fri 2024-12-06 20:21:39 UTC; 1min 4s ago
       Docs: man:sackd(8)
   Main PID: 2753 (sackd)
      Tasks: 11
     Memory: 900.0K (peak: 2.5M)
        CPU: 9ms
     CGroup: /system.slice/sackd.service
             └─2753 /usr/sbin/sackd --systemd --conf-server sackd-playground:6817

Dec 06 20:21:39 sackd-playground systemd[1]: Starting sackd.service - Slurm auth and cred kiosk daemon...
Dec 06 20:21:39 sackd-playground sackd[2753]: sackd: warning: sackd running as slurm instead of SlurmUser(root)
Dec 06 20:21:39 sackd-playground sackd[2753]: sackd: running
Dec 06 20:21:39 sackd-playground systemd[1]: Started sackd.service - Slurm auth and cred kiosk daemon.
```

jamesbeedy:
+1 🔥

NucciTheBoss:
Epic! :shipit: 

One thing I'd like to do in the future is rework how our integration tests are run. While the unit tests help us cover a lot of ground, it would still be nice to ensure that everything is working as expected with the daemons. It'll take time, and probably a separate PR, but we could style the `slurm_ops` tests similar to the [Slurm snap gambol tests](https://github.com/charmed-hpc/slurm-snap/blob/main/tests/integration/configless-slurm/configless-slurm.yaml), but use `slurm_ops` to drive Slurm's operations.

PR #56, opened 2024-12-04
feat(slurm_ops): add `acct_gather.conf` configuration file editor to `SlurmctldManager`

NucciTheBoss:
This PR adds an attribute for managing the _acct_gather.conf_ configuration file to the `SlurmctldManager` class. This attribute can then be used to handle changes related to configuring an acct_gather plugin from within the slurmctld charm. Having the editor be available will be necessary for re-enabling InfluxDB support in the Slurm charms.

Bumps the version of `slurm_ops` to 0.9.

### DIFF
--- a/tests/unit/test_slurm_ops.py
+++ b/tests/unit/test_slurm_ops.py
@@ -165,6 +165,27 @@ DownNodes=juju-c9fc6f-5 State=DOWN Reason="Maintenance Mode"
 PartitionName=DEFAULT MaxTime=30 MaxNodes=10 State=UP
 PartitionName=batch Nodes=juju-c9fc6f-2,juju-c9fc6f-3,juju-c9fc6f-4,juju-c9fc6f-5 MinNodes=4 MaxTime=120 AllowGroups=admin
 """
+EXAMPLE_ACCT_GATHER_CONFIG = """#
+# `acct_gather.conf` file generated at 2024-09-18 15:10:44.652017 by slurmutils.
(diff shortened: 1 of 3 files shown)


tests/unit/test_slurm_ops.py
```diff
@@ -625,6 +646,45 @@ def test_config(self, *_) -> None:
...
+
+    def setUp(self) -> None:
+        self.setUpPyfakefs()
+        self.manager = SlurmctldManager(snap=True)
+        self.fs.create_file(
+            "/var/snap/slurm/common/etc/slurm/acct_gather.conf",
+            contents=EXAMPLE_ACCT_GATHER_CONFIG,
+        )
+
+    def test_config(self) -> None:
```
dsloanm, line 660: Might be cleaner to split this into multiple tests. There's a lot of asserts in this function.

dsloanm (approved):
LGTM

tests/unit/test_slurm_ops.py
NucciTheBoss: I have it like this as slurmutils should be responsible for testing the nitty gritty of each editor/model with individual test functions. These unit tests are more or less for testing that we don't break the API for managing Slurm configs with slurm_ops.

If anything, we could probably make the tests shorter. The only thing is that we still have to be clever with mocking the uid and gid of the file as pyfakes doesn't provide any abstractions for doing that.

PR #58, opened 2024-12-11
feat(slurm_ops): add gres.conf configuration file editor to `SlurmctldManager`

NucciTheBoss:
This PR adds a gres.conf editor to `SlurmctldManager` so that we can configure GPUs in Charmed HPC through the gres.conf file. Generic resources need to be declared in gres.conf so that we can then declare nodes that have these resources within slurm.conf.

### Misc.

1. I took some time to clean up the `slurm_ops` unit tests by moving test cases into their own files. _test_slurm_ops.py_ was starting to become quite big so I broke tests down by focus area. Operations, package management, configuration management, etc.
2. I reorganized to `_ConfigManager` child classes to be in alphabetical order, so it makes the git diff appear larger that it actually is :sweat_smile:


jedel1043 (approved):
Looks good. Also, loving the new structure of the tests!

PR #60, opened 2025-01-07
chore(slurm_ops) bump `slurmutils` to `<1.0.0,>=0.11.0`

dsloanm:
Updates `PYDEPS` in `slurm_ops` from `"slurmutils~=0.10.0"` to `"slurmutils<1.0.0,>=0.11.0"` so latest `slurmutils` in the 0.x version series is used.

Follows discussion around version bumps needed for the latest GRES file editor.

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -116,7 +116,7 @@ PYDEPS = [
     "cryptography~=44.0.0",
     "pyyaml>=6.0.2",
     "python-dotenv~=1.0.1",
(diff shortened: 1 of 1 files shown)


NucciTheBoss (approved):
LGTM! :shipit: 

Only thing to note is that you may need to loosen the version of `slurmutils` within the _requirements.txt_ file of each Slurm charm. The requirements declaration in those files will take precedence over `PYDEPS`, so if you have a declaration such as `slurmutils ~= 0.8.0`, `slurmutils` version 0.8.* will be installed rather than the latest version of slurmutils released to PyPI.

PR #63, opened 2025-01-27
fix(slurm_ops): install `slurm-client` package on slurmd nodes

### LINKED ISSUE #62: `slurm-client` pkg missing on slurmd nodes
I can no longer run job scripts due to the missing `slurm-client` package on the slurmd nodes. Please add the slurm-client package back so that I can use sbatch and srun.

This issue presents itself when you submit a job-script via `sbatch` that calls `srun` (as the batch script itself and thus also `srun`, execute on the slurmd node).

`$ cat mysbatch.sh`

```bash
#!/bin/bash
#SBATCH -pslurmd

srun /home/ubuntu/myscript.sh
```

`sbatch mysbatch.sh` - this will fail with `/var/lib/slurm/slurmd/job00017/slurm_script: line 5: srun: command not found` because the batch script (`mysbatch.sh`) executes on the slurmd node.

The `slurm-client` package must exist on the slurmd nodes in order to submit jobs to the cluster- please put it back! Thanks!

NucciTheBoss:
This PR adds the `slurm-client` package to the installed package set for the `SlurmdManager` class. Installing this package on slurmd nodes will enable cluster users and admins to call common commands such as `srun` and `sbatch` from within their workloads.

### Related issues

* Fixes #62 

### Misc.

* Renames the `static` environment in _tox.ini_ to `typecheck`. Brings it forward for our standards for other repositories.

### DIFF
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -109,7 +109,7 @@ LIBAPI = 0
 
 # Increment this PATCH version before using `charmcraft publish-lib` or reset
 # to 0 if you are raising the major API version
-LIBPATCH = 12
+LIBPATCH = 13
 
 # Charm library dependencies to fetch during `charmcraft pack`.
 PYDEPS = [
--- a/lib/charms/hpc_libs/v0/slurm_ops.py
+++ b/lib/charms/hpc_libs/v0/slurm_ops.py
@@ -1,4 +1,4 @@
-# Copyright 2024 Canonical Ltd.
+# Copyright 2024-2025 Canonical Ltd.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


NucciTheBoss:
Hmm... integration tests pass locally. _Will dig deeper since this seems to be a recurrent problem._

jedel1043:
@NucciTheBoss The root cause sounds like the the same as https://github.com/charmed-hpc/slurm-charms/issues/67.

NucciTheBoss:
> @NucciTheBoss The root cause sounds like the the same as [charmed-hpc/slurm-charms#67](https://github.com/charmed-hpc/slurm-charms/issues/67).

Nah, I think the issue is likely with how we're setting up the integration tests. We might be getting a false positive here. Reason saying is that in the Slurm snap integration tests we're setting `StateSaveLocation` to _/var/lib/slurm/checkpoint_, however, that directory isn't created by the snap. slurmctld will fail with a fatal error if it cannot create and/or modify that directory.

dsloanm (approved):
Looks good, and nice catch with the integration tests.

PR #67, opened 2025-02-07
mirror the systemctl api via services api in the ServiceManager

jamesbeedy:
These changes tightly couple the systemctl api and the ServiceManager API, stop, start, enable, disable, restart, active - all do the same thing we would expect from systemctl by removing the `--now` flag and making the commands map to the simple systemctl commands without additional arguments.


jedel1043:
Closing since #68 was merged.

PR #70, opened 2025-02-07
fix(slurm_ops): remove `--now` flag from enable/disable

NucciTheBoss:
This PR removes the `--now` flag from `_SystemctlServiceManager` and enables static type checking in our CI pipeline for pull requests. As part of improving the static type checking, I also added some required type annotations.


jamesbeedy (approved):
+1

PR #71, opened 2025-02-11
docs: add SECURITY.md

jedel1043:
Adds a simple SECURITY.md about our security policy for the repository.

Note that reporting vulnerabilities for this repo is currently disabled, so we'll need to enable that.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,32 @@
+# Security policy
+
+## Supported versions
+
+Security updates for charm libraries will be released for the latest two major versions released on Charmhub.
+
+## Reporting a vulnerability
+
+Please provide a description of the issue, the steps you took to create the issue, affected versions, and, if known,
+mitigations for the issue.
+
+The preferred way to report a security issue is through [GitHub Security Advisories][gsa]. See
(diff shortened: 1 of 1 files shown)


SECURITY.md
```diff
@@ -0,0 +1,33 @@
+# Security policy
+
+## Supported versions
+
+Security updates will be released after vulnerabilities have been identified and fixed. At this point in time,
+we will only release security patches for charm libraries to their latest major versions released on Charmhub.
```
NucciTheBoss: Similar to the `slurmutils` _SECURITY.md_, support latest two major versions, but scoped to only the charm libraries released to Charmhub:

```suggestion
Security updates for charm libraries will be released for the latest two major versions released on Charmhub.
```

Note that this might change in the future depending CharmTechs work on Charm SDKs to get rid of complex charm libraries that rely on PYDEPS.

NucciTheBoss (changes requested):
Suggested revision to make the `Supported versions` policy similar to the one for `slurmutils`.

NucciTheBoss (approved):
LGTM! :shipit:

PR #81, opened 2025-04-08
fix: enable automatic restart of slurmctld in case of failure

jedel1043:
# Pre-submission checklist

 * [x] I have insured that lint, typecheck, and unit tests complete successfully.

## Summary of changes

Enable automatic restart of slurmctld on its systemd service file in case of failure, which
could happen if slurmctld tries to query slurmdbd while its backing database is not yet ready.

## Docs

Only internal change.


dsloanm (approved):
Looks good - should make `slurmctld` more resilient in general.

NucciTheBoss (approved):
LGTM! It will be good to leverage `systemd`'s built in capability to make `slurmctld` more resilient within the charm.

PR #110, opened 2025-09-01
fix: ignore `relation-broken` when a unit is removed

dsloanm:
Prevents erroneous revocation of secrets and clearing of relation data when an application leader is removed during scale down.

# Pre-submission checklist

 * [X] I read and followed the CONTRIBUTING guidelines.
 * [X] I have ensured that lint, typecheck, and unit tests complete successfully.

## Summary of changes

This PR provides a fix for https://bugs.launchpad.net/juju/+bug/1979811.

Specifically, `relation-broken` events do not have enough information to determine whether they have been fired because the current unit is being removed, or because the integration itself is being removed. In the case where only the current unit is being removed, relation data should not be cleared nor secrets revoked. However, this is happening in the latest version of `hpc-libs`, resulting in application failure when a unit leader is removed during scale down.

As a workaround, `StoredState` is introduced to allow a flag to be set in the `relation-departed` hook if the current unit is being removed. The flag is then checked in the `relation-broken` event handler to determine whether data should be cleared or not.

`StoredState` is used rather than the unit databag to prevent extraneous `relation-changed` events from firing as the unit is being removed.

As an alternative, a `planned_units()` check in `relation-deparated` was considered, that would clear relation data when the last unit departed, but was discarded as scaling the application to 0 (without breaking the relation) then attempting to scale up would cause failure.

This fix is expected to be temporary as the introduction of `StoredState` is undesirable and the root issue may be resolved by Juju 4. The fix should be reassessed when the code is migrated in future.

This solution is inspired by:
https://github.com/canonical/postgresql-operator/blob/e13b871/src/relations/db.py#L193

This is necessary to support the merging of high-availability functionality for the `slurmctld` charm.

## Docs

* [ ] I have created a pull request to add or update relevant documentation in [charmed-hpc/docs](https://github.com/charmed-hpc/docs) or another documentation location.

[//]: # (If documentation has been updated or added in a location other than charmed-hpc/docs, please note the location here.)

Or:

* [X] I confirm that this pull request requires no changes or additions to documentation.

This is a bug fix with no user-facing changes.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "hpc-libs"
-version = "0.1.0"
+version = "0.1.1"
 requires-python = ">=3.10"
 description = "Collection of utilities to manage HPC related services."
 authors = [
--- a/src/hpc_libs/interfaces/base.py
+++ b/src/hpc_libs/interfaces/base.py
@@ -88,6 +88,25 @@ class Interface(ops.Object):
           a requirer/provider to successfully process a `RelationChangedEvent`.
     """
 
+    # A temporary fix for:
+    # https://bugs.launchpad.net/juju/+bug/1979811
+    #
+    # `relation-broken` events do not have enough information to determine whether they have been
+    # fired because the current unit is being removed, or because the integration itself is being
+    # removed. In the case where only the current unit is being removed, relation data should not be
+    # cleared nor secrets revoked.
+    #
+    # As a workaround, a StoredState flag is set in the `relation-departed` hook if the current unit
+    # is being removed. The flag is then checked in the `relation-broken` event handler to determine
+    # whether data should be cleared or not.
+    #
+    # StoredState is used rather than the unit databag to prevent extraneous `relation-changed`
+    # events from firing as the unit is being removed.
+    #
+    # Inspired by:
+    # https://github.com/canonical/postgresql-operator/blob/e13b871/src/relations/db.py#L193
+    stored_state = ops.StoredState()
+
     def __init__(
         self,
         charm: ops.CharmBase,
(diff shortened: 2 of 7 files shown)


jedel1043:
Thought: I'm really wondering if we have the correct design for the secrets here.

Looking at the code, it seems that we're creating one secret per relation for each key, which is causing these kind of problems because we need to delete all the unused secrets after a relation gets removed. However, according to [Juju's documentation](https://documentation.ubuntu.com/juju/3.6/reference/secret/#secret-lifecycle), Juju already keeps track of all observers for a secret, and triggers a `secret-remove` event when there are no more observers for a secret.

IMO what we should have instead is a global secret for the keys (or a secret per key, but maybe having two secrets would be nicer to manage), and we should modify `set_controller_data` so that we are able to call it without an `integration_id`. Since all secrets have labels, we could use a label to store the secret data, then retrieve it and grant permission to the secret when either `set_controller_data` gets called or the `relation-joined` event gets triggered. That way, we just need to handle `secret-removed` which has the semantics we want; it only gets triggered if it is safe to remove a revision of a secret.

We do something very similar to this on the [`filesystem-client`](https://github.com/charmed-hpc/filesystem-charms/blob/f2cfbbe4111bb265209dfb216982e1d25e5385f8/charms/filesystem-client/lib/charms/filesystem_client/v0/filesystem_info.py#L667-L685), which uses a peer relation databag to store the mount info instead, but I figured if what we're storing is just a bunch of keys, this could be a nicer way to fix the problem.

dsloanm:
> Thought: I'm really wondering if we have the correct design for the secrets here.

You make a good case here. We should probably review how we're using secrets but we'll still need this workaround for preserving general relation data. For example, we still want to prevent [`_on_slurmd_disconnected()`](https://github.com/charmed-hpc/slurm-charms/blob/main/charms/slurmctld/src/charm.py#L298) in `slurmctld` from erroneously deleting a compute node partition when the unit leader is removed. That's addressed in this PR by not emitting [`slurmd_disconnected`](https://github.com/charmed-hpc/hpc-libs/blob/587c01a96541b518385fa2d397d07719eb6a8784/src/hpc_libs/interfaces/slurm/slurmd.py#L186) if the current unit is departing.

As I was writing this comment, I see you've given a review 🎉! I was going to propose we open an issue to revisit the secret design - agreed it's something for the future.

NucciTheBoss:
Opened exploratory/enhancement request issue here: https://github.com/charmed-hpc/hpc-libs/issues/111

PR #128, opened 2026-04-14
Allow setting the app status with StopCharm

jedel1043:
# Pre-submission checklist

## Summary of changes

Adds a `app_status` field to `StopCharm` so that consumers can also set the app status if the current unit is the leader unit.

This is required in `filesystem_charms` such that we can reuse the `refresh` decorator there without having to create a separate handler.

## Docs

API docs for charmed-hpc-libs are not public.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "charmed-hpc-libs"
-version = "0.1.1"
+version = "0.1.2"
 requires-python = ">=3.12"
 description = "Collection of utilities to manage HPC related services."
 authors = [
--- a/src/charmed_hpc_libs/ops/conditions.py
+++ b/src/charmed_hpc_libs/ops/conditions.py
@@ -48,12 +48,21 @@ type Condition[T: ops.CharmBase] = Callable[[T], ConditionEvaluation]
 
 
 class StopCharm(Exception):  # noqa N818
-    """Exception raised to set high-priority status message."""
+    """Exception raised to set high-priority status message.
 
-    @property
-    def status(self) -> ops.StatusBase:
-        """Get charm status passed as the first argument to this exception."""
-        return self.args[0]
+    Args:
+        status: The status to set on the unit (and optionally the application).
+        app_status: If ``True``, also set the application status to the unit
+            status when the unit is the leader.
+    """
+
+    def __init__(self, status: ops.StatusBase, *, set_app_status: bool = False) -> None:
+        super().__init__(status)  # only return the `status` message when using `str(e)`
+        self.status = status
+        self.set_app_status = set_app_status
+
+    def __repr__(self) -> str:  # noqa D105
+        return f"StopCharm({self.status!r}, set_app_status={self.set_app_status!r})"
 
 
 def refresh[T: ops.CharmBase](hook: Callable[[T], ops.StatusBase] | None = None) -> Callable:
@@ -74,7 +83,7 @@ def refresh[T: ops.CharmBase](hook: Callable[[T], ops.StatusBase] | None = None)
                 _logger.debug(
                     (
                         "`StopCharm` exception raised while running `%s` event handler `%s.%s` ",
-                        "on unit '%s'. setting status to `%s`",
+                        "on unit '%s'. setting unit status to `%s`",
                     ),
                     event.__class__.__name__,
                     charm.__class__.__name__,
@@ -83,6 +92,15 @@ def refresh[T: ops.CharmBase](hook: Callable[[T], ops.StatusBase] | None = None)
                     e.status,
                 )
                 charm.unit.status = e.status
+                if e.set_app_status:
+                    try:
+                        charm.app.status = e.status  # type: ignore
+                        _logger.debug(
+                            "setting app status to `%s`",
+                            e.status,
+                        )
+                    except RuntimeError:
+                        pass
                 return
 
             if hook is not None:
--- a/tests/unit/ops/test_conditions.py
+++ b/tests/unit/ops/test_conditions.py
@@ -20,6 +20,7 @@ from ops import testing
 
 from charmed_hpc_libs.ops import (
     ConditionEvaluation,
+    StopCharm,
     block_unless,
     integration_exists,
     integration_not_exists,
@@ -134,3 +135,45 @@ def test_refresh(mock_charm) -> None:
     )
 
     assert state.unit_status == ops.BlockedStatus("Waiting for integrations: [`metrics-server`]")
+
+
+class AppStatusCharm(ops.CharmBase):
+    """Mock charm for testing `StopCharm` app status propagation."""
+
+    def __init__(self, framework: ops.Framework) -> None:
+        super().__init__(framework)
+        framework.observe(self.on.install, self._on_install)
+        framework.observe(self.on.start, self._on_start)
+
+    @refresh(hook=None)
+    def _on_install(self, _: ops.InstallEvent) -> None:
+        raise StopCharm(ops.BlockedStatus("blocked for app"), set_app_status=True)
+
+    @refresh(hook=None)
+    def _on_start(self, _: ops.StartEvent) -> None:
+        raise StopCharm(ops.BlockedStatus("blocked for unit only"))
+
+
+@pytest.fixture(scope="function")
+def app_status_charm() -> testing.Context[AppStatusCharm]:
+    return testing.Context(AppStatusCharm, meta={"name": "app-status-charm"})
+
+
+@pytest.mark.parametrize("is_leader", (True, False))
+def test_refresh_app_status(app_status_charm, is_leader) -> None:
+    """Test that `refresh` sets app status with `set_app_status=True` only when unit is leader."""
+    state = app_status_charm.run(
+        app_status_charm.on.install(), state=testing.State(leader=is_leader)
+    )
+    assert state.unit_status == ops.BlockedStatus("blocked for app")
+    if is_leader:
+        assert state.app_status == ops.BlockedStatus("blocked for app")
+    else:
+        assert state.app_status == ops.UnknownStatus()
+
+
+def test_refresh_no_app_status_by_default(app_status_charm) -> None:
+    """Test that `refresh` does not touch app status when `set_app_status` is `False` (default)."""
+    state = app_status_charm.run(app_status_charm.on.start(), state=testing.State(leader=True))
+    assert state.unit_status == ops.BlockedStatus("blocked for unit only")
+    assert state.app_status == ops.UnknownStatus()
--- a/uv.lock
+++ b/uv.lock
@@ -13,7 +13,7 @@ wheels = [
 
 [[package]]
 name = "charmed-hpc-libs"
-version = "0.1.3"
+version = "0.1.2"
 source = { editable = "." }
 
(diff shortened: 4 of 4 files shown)


src/charmed_hpc_libs/ops/conditions.py
```diff
@@ -48,13 +48,33 @@ class ConditionEvaluation(NamedTuple):  # noqa D101
...
+    """
+
+    def __init__(
+        self, status: ops.StatusBase, *, app_status: ops.StatusBase | None = None
+    ) -> None:
+        # passing `app_status` here ensures the app status is also stored
+        # We could also add an `app_status` and manually implement `__str__`, but
+        # this exception is supposed to be caught by condition handlers, so we
+        # kinda complicate the code for no real reason.
+        super().__init__(status, app_status)
```
NucciTheBoss: Using this approach feels a bit clunky as we'll always need to set a status for the unit when we might just want to only update the app status. We should try to avoid visual clutter where possible in the `juju status` since there's limited real estate for grabbing admins' attention.

What if we made `app_status` a boolean instead and changed the evaluation order for whether to update the unit's or application's status. E.g. something like:

```python
if self.app_status:
    # Ask for forgiveness instead of permission.
    try:
        charm.app.status = self.status
    except RuntimeError:
        pass  # Do nothing
else:
    charm.unit.status = self.status

return
```

With this suggested approach, if there is a status we need to set at the application level, but we also want to set a unit status, we can just set the unit status before raising `StopCharm`. The "check" hook that is passed to `refresh` won't be executed if a `StopCharm` exception is raised in an event handler:

```python
try:
    ...  # Do something
except:
    self.unit.status = ops.BlockedStatus("I'm the unit's that's cooked")
    raise StopCharm(
        ops.BlockedStatus("Some unit in this application is broke, busted, and disgusted"),
        app_status=True,
    )
```

I think this approach is  cleaner than using two independent status messages to maintain backwards compatibility with existing uses of `StopCharm` and `refresh` in our charms. What are your thoughts?

pyproject.toml
```diff
@@ -18,7 +18,7 @@ build-backend = "hatchling.build"
 
 [project]
 name = "charmed-hpc-libs"
-version = "0.1.1"
+version = "0.1.3"
```
NucciTheBoss: [nit]: what happened to `0.1.2`??

src/charmed_hpc_libs/ops/conditions.py
```diff
@@ -48,13 +48,33 @@ class ConditionEvaluation(NamedTuple):  # noqa D101
...
+            Defaults to ``None``, which skips setting the app status.
+    """
+
+    def __init__(
+        self, status: ops.StatusBase, *, app_status: ops.StatusBase | None = None
+    ) -> None:
+        # passing `app_status` here ensures the app status is also stored
+        # We could also add an `app_status` and manually implement `__str__`, but
+        # this exception is supposed to be caught by condition handlers, so we
+        # kinda complicate the code for no real reason.
```
NucciTheBoss: We could also use protected attributes to store the value of `status` and `app_status` like how the snap charm library does it: https://github.com/canonical/charmlibs/blob/dae489f3d88e41650f93fe714d2921beee7452db/snap/src/charmlibs/snap/_snap.py#L181-L210

That way we don't have to rely on the ordering of positional arguments always being the same.

NucciTheBoss (changes requested):
Some thoughts around the implementation of how we're setting the app status using `StopCharm`. I'm worried about `StopCharm` bloating the output of `juju status` when we really only want to set the application status messages. _Something something less is more._

Let me know if you have any questions before diving into my suggestions!

src/charmed_hpc_libs/ops/conditions.py
jedel1043: The rationale for doing it this way is that you definitely wanna set the unit status if you're setting the app status. Otherwise, you could get in a situation where the app status would show "blocking" but all units would show "active", which is very confusing.

Though TBF it is kinda a bad API since you could set the unit and app to different status types, so what do you think about just passing a `set_app_status` boolean instead? Then if the unit is leader and the boolean is set, we also set the app status (probably doing what you mention about trying to set it and ignoring errors if the unit is not leader)

pyproject.toml
jedel1043: whoops, I accidentally bumped the version twice haha

src/charmed_hpc_libs/ops/conditions.py
NucciTheBoss: `set_app_status` - or just `app_status` with a docstring that explains that this will update the app status too with the same message as the unit status - sounds agreeable to me. That way all units can have the same failure message, and the application status reflects that as well.

Although IIRC I remember reading somewhere that if an application doesn't have an app status set then by default the status of the leader unit in the application will become the app status. You might want to double check if we can't just get away with updating the leader unit's status to update the application status. That said, I do like your `set_app_status` suggestion if I'm misremembering how unit statuses work :sweat_smile:

jedel1043: Ah yeah, that's the "normal" behaviour, but for subordinate charms we kinda require having different statuses for apps and units, and setting the app status "removes" the automatic fallback to the leader unit.

For example, on the filesystem client we specifically need to set the app status message to something different because each subordinate could integrate with a filesystem consumer using a different path for each application, and the units report the current path the filesystem is mounted on. I guess we could remove this custom status message to allow status fallback, but it might look weird to have a slightly imprecise `mounted on /path` status message for the whole app.

src/charmed_hpc_libs/ops/conditions.py
jedel1043: Hmm yeah, that might be better. If we go with `set_app_status` I'll also migrate the error to this design.

src/charmed_hpc_libs/ops/conditions.py
NucciTheBoss: Hmm... maybe a radical suggestions here, but do we really need every filesystem-client unit to have it's status set to "<filesystem> filesystem mounted at /<mountpoint>"?

My reasoning against the unit status message for filesystem-client charm is that I can see the unit status message becoming quite noisy when we scale up partitions to more than just couple of compute nodes. The filesystem-client units being in `active` status and the Juju agents being `idle` are enough to imply to me that everything is operating as expected. For example, all the Slurm charms have a blank status message when everything is operating as expected (except for slurmctld in HA configuration). If a cluster administrator forgets the mount path the filesystem-client application is managing they can either:

1. Choose a name for their filesystem-client application that maps to the name of the mount point that it's managing. E.g. `scratch` for `scratch/`, `work` for `work/`, etc.
2. View the output of `juju config ...` to see what the configured mount point is. E.g. `juju config filesystem-client mountpoint`.

With this suggested approach, we could possibly solve two problems at once:

1. We reduce the amount of noise filesystem-client units add to the `juju status` output.
2. We only have to update the status of individual units if they are degraded. Less noise in `juju status` also makes it easier to quickly identify problematic units at a glance.
3. We only have to update the app status when conditions are good. For example, set the app status to "<filesystem> filesystem mounted at /<mountpoint>" after the leader successfully mounts the shared filesystem.

This way we don't need to chop up `StopCharm` so that it doubles for setting both unit and application status with the same message (error messages can be easily chased down to the problematic unit), and we can reduce the amount of noise filesystem-client adds to the output of `juju status`.

Let me know your thoughts here.

jedel1043: For getting the mount path:
1 gets kinda annoying if the path is not simple, and doesn't work for using the filesystem client with more than 1 application (one app could require mounting to `/srv` and the other to `/export` or something)

2 is very annoying to do if you're integrating using the `mount` relation instead of the `juju-info` relation, since the path is only visible from the relation databag. Showing that info on the status was the most straightforward way of showing that everything is working correctly aside from looking at the output of `juju debug-log`.

NucciTheBoss: Chatted during our 1:1 today and decided the best path forward to keep status messages in `filesystem-client` as they are, and use the `set_app_status` boolean to indicate that the application status should be updated as well. 

Figuring out a potentially better way to communicate all the endpoints managed by a particular `filesystem-client` unit will be shelved for next cycle since (a) we need more people to give us usability feedback and (b) we need to research more approaches before jumping on one particular implementation. I quite like the `show-proxied-endpoints` action from the `traefik-k8s` operator, but we need to determine if it's relevant for our use case during a formal spike

I'm going to mark this conversation and wait for the latest changes since this convo has gotten quite long :sweat_smile:

pyproject.toml
jedel1043: Fixed on the latest rebase.

src/charmed_hpc_libs/ops/conditions.py
jedel1043: Addressed on the latest rebase.

src/charmed_hpc_libs/ops/conditions.py
```diff
@@ -83,6 +93,19 @@ def wrapper(charm: T, *args: ops.EventBase, **kwargs: Any) -> None:
...
+                        charm.app.status = e.status  # type: ignore
+                        _logger.debug(
+                            "setting app status to `%s`",
+                            e.status,
+                        )
+                    except RuntimeError:
+                        # According to https://documentation.ubuntu.com/ops/latest/reference/ops/#ops.Application.status,
+                        # ops will raise a RuntimeError if the unit trying to set
+                        # the app status is not the leader. We can ignore
+                        # the error in that case.
```
NucciTheBoss: [nit]: I don't think this comment is necessary. The docstring for `charm.app.status` already mentions that `RuntimeError` is raised if a non-leader unit attempts to set the application status, so whoever is reading this block of code can infer that we're silently ignoring `RuntimeError` because a non-leader unit attempted to set the application status.

```suggestion
```

If you think we should still have the comment, we could shorten it to something like `A RuntimeError is raised if an non-leader unit tries to update the application's status`, but imo I think a charm developer can infer what's going on here by reading existing documentation.

src/charmed_hpc_libs/ops/conditions.py
```diff
@@ -48,12 +48,22 @@ class ConditionEvaluation(NamedTuple):  # noqa D101
...
+            status when the unit is the leader.
+    """
+
+    def __init__(self, status: ops.StatusBase, *, set_app_status: bool = False) -> None:
+        super().__init__(status)  # only return the `status` message when using `str(e)`
+        self.status = status
+        self.set_app_status = set_app_status
+
+    def __repr__(self) -> str:
+        """Represent the StopCharm class."""
```
NucciTheBoss: [nit]: you can just use `# noqa D105` here. It drives me crazy that `ruff` expects us to add docstrings to magic methods like `__repr__` and `__add__` when their behavior is already well known. It's on the same level to me as comments like `print("hello world")  # prints the string hello world` 😅 

```suggestion
    def __repr__(self) -> str:  # noqa D105
```

NucciTheBoss (changes requested):
Two tiny nits and then I'm ready to merge this PR 👍

src/charmed_hpc_libs/ops/conditions.py
jedel1043: Removed the doc in favour of the noqa

NucciTheBoss (approved):
LGTM! :shipit: 

Nice work!

PR #129, opened 2026-04-14
feat: add `NodeExporterManager` and `exporters` module

### LINKED ISSUE #127: Add `node_exporter` support
We should add support for [`node_exporter`](https://github.com/prometheus/node_exporter) to `charmed-hpc-libs`. While the OpenTelemetry Collector operator includes a `node_exporter` instance by default, it doesn't make sense to use the OpenTelemetry Collector operator to provide `node_exporter` on potentially massive applications like `slurmd`. It doesn't make sense because OpenTelemetry Collector relies on a peer integration for replication, and peer integrations in Juju do not scale on large applications.

Because of these scalability issues, `charmed-hpc-libs` should include code for managing a `node_exporter` installation on charms that require it. It can be styled after the original `PrometheusExporterManager` from `slurm-ops` where it implements aspects of `OpsManager` and `ServiceManager` to encapsulate the management logic around the `node_exporter` service. Several steps will need to be completed as part of this work:

1. A base class for operations managers will need to be defined. It should be simpler than the operations managers in `slurm-ops` since we won't be aiming for compatibility between two different styles of packaging Slurm.
2. A snap operations manager will need to be defined to facilitate connections between `snapd` and `node_exporter`
3. A top-level _exporters.py_ module will need to be created. This is where common exporters will be found that can be installed on a broad variety of charms.
4. Create the `NodeExporterManager` class that will encapsulate all the operations logic for the `node_exporter` service.

NucciTheBoss:
# Pre-submission checklist

## Summary of changes

This PR proposes adding a `NodeExporterManager` class that can be used to manage a [`node_exporter`](https://snapcraft.io/node-exporter) service on machine charms. This class can be used to enable exporting machine metrics without the OpenTelemetry Collector operator.

Key changes include:

1. Adding the protocol `OperationsManager`. This class is inspired from the `OpsManager` class in `slurm-ops`, but more generic.
2. Adding the `SnapOperationsManager` class. This class provides method for managing common operations around a snap package such as installation and removal, connecting plugs to different slots, and managing the packages; configuration.
3. Adding the `SnapLifecycleManager` class. This class inherits from both `SnapServiceManager` and `SnapOperationsManager`. This class is intended to be used for building classes that are intended to manage the entire lifecycle of an application, unlike `slurm-ops` where we use composition instead to create managers for more complicated services.
4. Adding the `NodeExporterManager` class. This class is used to manage a snap-based node exporter service on machine charms. It can be used in charms like `sackd` and `slurmd` to provide a node exporter endpoint but not require an OpenTelemetry Collector instance. 

- Closes https://github.com/charmed-hpc/hpc-libs/issues/127

### Other things to note

 I did a preliminary co-pilot review in my IDE (PyCharm).

## Docs

* [ ] I have created a pull request to add or update relevant documentation in [charmed-hpc/docs](https://github.com/charmed-hpc/docs) or another documentation location.

[//]: # (If documentation has been updated or added in a location other than charmed-hpc/docs, please note the location here.)

Or:

This is a non-user facing change.

### DIFF
--- a/src/charmed_hpc_libs/ops/exporters.py
+++ b/src/charmed_hpc_libs/ops/exporters.py
@@ -38,7 +38,7 @@ class NodeExporterManager(SnapLifecycleManager):
             >>> ["ntp"]
         """
         try:
-            return self.get("collectors").split()
+            return self.config.get("collectors").split()
         except SnapError:
             # `SnapError` is raised if `collectors` option is empty.
             return []
@@ -54,9 +54,9 @@ class NodeExporterManager(SnapLifecycleManager):
             - The service will restart after the list of collectors is updated.
         """
         if len(value) == 0:
-            self.unset("collectors")
+            self.config.unset("collectors")
         else:
-            self.set({"collectors": " ".join(value)})
+            self.config.set({"collectors": " ".join(value)})
 
     def get_no_collectors(self) -> list[str]:
         """Get the list of disabled collectors.
@@ -68,7 +68,7 @@ class NodeExporterManager(SnapLifecycleManager):
             >>> ["mdadm", "netstat"]
         """
         try:
-            return self.get("no-collectors").split()
+            return self.config.get("no-collectors").split()
         except SnapError:
             # `SnapError` is raised if `no-collectors` option is empty.
             return []
@@ -84,9 +84,9 @@ class NodeExporterManager(SnapLifecycleManager):
             - The service will restart after the list of disabled collectors is updated.
         """
         if len(value) == 0:
-            self.unset("no-collectors")
+            self.config.unset("no-collectors")
         else:
-            self.set({"no-collectors": " ".join(value)})
+            self.config.set({"no-collectors": " ".join(value)})
 
     def get_web_listen_address(self) -> str:
         """Get web address and port used by the `node-exporter` service.
@@ -100,7 +100,7 @@ class NodeExporterManager(SnapLifecycleManager):
             >>> "127.0.0.1:9200"
         """
         try:
-            return self.get("web.listen-address")
+            return self.config.get("web.listen-address")
         except SnapError:
             # `SnapError` is raised if `web.listen-address` option is empty.
             return ""
@@ -116,6 +116,6 @@ class NodeExporterManager(SnapLifecycleManager):
             - The service will restart after the web listen address is updated.
         """
         if value:
-            self.set({"web.listen-address": value})
+            self.config.set({"web.listen-address": value})
         else:
-            self.unset("web.listen-address")
+            self.config.unset("web.listen-address")
--- a/src/charmed_hpc_libs/ops/exporters.py
+++ b/src/charmed_hpc_libs/ops/exporters.py
@@ -28,16 +28,13 @@ class NodeExporterManager(SnapLifecycleManager):
     def __init__(self) -> None:
         super().__init__("node-exporter")
 
-    @property
-    def collectors(self) -> list[str]:
+    def get_collectors(self) -> list[str]:
         """Get the list of optionally-enabled collectors.
-
-        Providing an empty list `[]` will disable all optionally-enabled collectors.
 
         Examples:
             >>> node_exporter = NodeExporterManager()
-            >>> node_exporter.collectors = ["ntp"]
-            >>> node_exporter.collectors
+            >>> node_exporter.set_collectors(["ntp"])
+            >>> node_exporter.get_collectors()
             >>> ["ntp"]
         """
         try:
@@ -46,21 +43,28 @@ class NodeExporterManager(SnapLifecycleManager):
             # `SnapError` is raised if `collectors` option is empty.
             return []
 
-    @collectors.setter
-    def collectors(self, value: Collection[str]) -> None:
+    def set_collectors(self, value: Collection[str]) -> None:
+        """Enable optional collectors.
+
+        Args:
+            value: Collection of optional collectors to enable.
+
+        Notes:
+            - Providing an empty collection will unset all optionally-enabled collectors.
+            - The service will restart after the list of collectors is updated.
+        """
         if len(value) == 0:
             self.unset("collectors")
         else:
             self.set({"collectors": " ".join(value)})
 
-    @property
-    def no_collectors(self) -> list[str]:
+    def get_no_collectors(self) -> list[str]:
         """Get the list of disabled collectors.
 
         Examples:
             >>> node_exporter = NodeExporterManager()
-            >>> node_exporter.no_collectors = ["mdadm", "netstat"]
-            >>> node_exporter.no_collectors
+            >>> node_exporter.set_no_collectors(["mdadm", "netstat"])
+            >>> node_exporter.get_no_collectors()
             >>> ["mdadm", "netstat"]
         """
         try:
@@ -69,23 +73,30 @@ class NodeExporterManager(SnapLifecycleManager):
             # `SnapError` is raised if `no-collectors` option is empty.
             return []
 
-    @no_collectors.setter
-    def no_collectors(self, value: Collection[str]) -> None:
+    def set_no_collectors(self, value: Collection[str]) -> None:
+        """Disable collectors.
+
+        Args:
+            value: Collection of collectors to disable.
+
+        Notes:
+            - Providing an empty collection will unset all disabled collectors.
+            - The service will restart after the list of disabled collectors is updated.
+        """
         if len(value) == 0:
             self.unset("no-collectors")
         else:
             self.set({"no-collectors": " ".join(value)})
 
-    @property
-    def web_listen_address(self) -> str:
+    def get_web_listen_address(self) -> str:
         """Get web address and port used by the `node-exporter` service.
 
         Examples:
             >>> node_exporter = NodeExporterManager()
-            >>> node_exporter.web_listen_address
+            >>> node_exporter.get_web_listen_address()
             >>> ":9200"
-            >>> node_exporter.web_listen_address = "127.0.0.1:9200"
-            >>> node_exporter.web_listen_address
+            >>> node_exporter.set_web_listen_address("127.0.0.1:9200")
+            >>> node_exporter.get_web_listen_address()
             >>> "127.0.0.1:9200"
         """
         try:
@@ -94,10 +105,17 @@ class NodeExporterManager(SnapLifecycleManager):
             # `SnapError` is raised if `web.listen-address` option is empty.
             return ""
 
-    @web_listen_address.setter
-    def web_listen_address(self, value: str) -> None:
-        self.set({"web.listen-address": value})
+    def set_web_listen_address(self, value: str) -> None:
+        """Set web address and port used by the `node-exporter` service.
 
-    @web_listen_address.deleter
-    def web_listen_address(self) -> None:
-        self.unset("web.listen-address")
+        Args:
+            value: The web listen address to use for the `node-exporter` service.
+
+        Notes:
+            - Providing an empty string will unset the web listen address.
+            - The service will restart after the web listen address is updated.
+        """
+        if value:
+            self.set({"web.listen-address": value})
+        else:
+            self.unset("web.listen-address")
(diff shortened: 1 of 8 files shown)


src/charmed_hpc_libs/ops/machine/snap.py
```diff
@@ -48,6 +50,85 @@ def snap(*args: str, **kwargs: Any) -> tuple[str, int]:  # noqa D417
...
+        """
+        result = snap("get", "-d", self._snap, key)
+        try:
+            data = json.loads(result[0])
+        except json.JSONDecodeError as e:
+            raise SnapError(
+                f"Failed to decode value of configuration option '{key}' for snap '{self._snap}'"
+            ) from e
+
+        return data[key]
```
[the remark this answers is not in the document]
NucciTheBoss: This is incorrect. `snap get -d <snap_name> <key_name>` will fail with a non-zero exit code if the requested key is not present in the snap's configuration. Therefore, we can assume the key is present if a `SnapError` isn't raised.

src/charmed_hpc_libs/ops/machine/snap.py
```diff
@@ -48,6 +50,85 @@ def snap(*args: str, **kwargs: Any) -> tuple[str, int]:  # noqa D417
...
+            plug: The plug to connect.
+            service: The snap service name to plug into.
+            slot: The snap service slot to plug in to.
+        """
+        command = ["connect", f"{self._snap}:{plug}"]
+        if service and slot:
+            command.append(f"{service}:{slot}")
+        elif slot:
+            command.append(slot)
+
```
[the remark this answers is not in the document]
NucciTheBoss: `service` can be provided directly if the service has a slot that the snap can plug in to. I'll need to update this code. Also, `slot` is formatted incorrectly when it is appended to the list so I will fix that as well.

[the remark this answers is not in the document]
NucciTheBoss: Fixed.

src/charmed_hpc_libs/ops/machine/snap.py
```diff
@@ -48,6 +50,85 @@ def snap(*args: str, **kwargs: Any) -> tuple[str, int]:  # noqa D417
...
+            key: Snap configuration key to get the value of.
+
+        Examples:
+            >>> package = SnapOperationsManager("slurm")
+            >>> package.get("exporter.port")
+            >>> 9100
+        """
+        result = snap("get", "-d", self._snap, key)
+        try:
+            data = json.loads(result[0])
```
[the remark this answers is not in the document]
NucciTheBoss: `snap get -d` will either always return a valid JSON object containing the requested key or an error code. The `stdout=None` annotation for `call` exists because there are times that we do not wish to capture `stdout`. It's not because sometimes the output of commands will be blank.

src/charmed_hpc_libs/ops/exporters.py
```diff
@@ -0,0 +1,103 @@
...
+            >>> node_exporter = NodeExporterManager()
+            >>> node_exporter.no_collectors = ["mdadm", "netstat"]
+            >>> node_exporter.no_collectors
+            >>> ["mdadm", "netstat"]
+        """
+        try:
+            return self.get("no-collectors").split()
+        except SnapError:
+            # `SnapError` is raised if `no-collectors` option is empty.
+            return []
```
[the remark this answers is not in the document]
NucciTheBoss: Such is life with subprocess wrappers. While we could wire into the snapd REST API:

1. That's complicated.
2. That's beyond our needs.

We could add more granular alerts such as `SnapNotFoundError` or `SnapTimeoutError`, however, for the current needs of this PR, this should be enough. These properties will largely be used for retrieving config values for logging, so if there is a true problem with `snap`, these errors will be emitted by the `set`/`unset` methods. This is similar to how `jubilant` works where a generic `CLIError` is emitted if there are any problems with Juju.

That said, we can expect a `SnapError` to be raised if the configuration field is empty, so we should treat this default behavior by `snap` as if the `no_collectors` field is empty.

src/charmed_hpc_libs/ops/exporters.py
```diff
@@ -0,0 +1,103 @@
...
+            >>> node_exporter = NodeExporterManager()
+            >>> node_exporter.collectors = ["ntp"]
+            >>> node_exporter.collectors
+            >>> ["ntp"]
+        """
+        try:
+            return self.get("collectors").split()
+        except SnapError:
+            # `SnapError` is raised if `collectors` option is empty.
+            return []
```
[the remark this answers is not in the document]
NucciTheBoss: See my response on the `no_collectors` for why I don't think this is a huge issue.

src/charmed_hpc_libs/ops/exporters.py
```diff
@@ -0,0 +1,103 @@
...
+            >>> ":9200"
+            >>> node_exporter.web_listen_address = "127.0.0.1:9200"
+            >>> node_exporter.web_listen_address
+            >>> "127.0.0.1:9200"
+        """
+        try:
+            return self.get("web.listen-address")
+        except SnapError:
+            # `SnapError` is raised if `web.listen-address` option is empty.
+            return ""
```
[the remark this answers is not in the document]
NucciTheBoss: See my response on the `no_collectors` comment for why I don't think this is a huge issue.

I think a future improvement we could make is adding more granular errors such as `SnapTimeoutError` or `SnapNotFoundError`, but I think these changes are out of scope for this PR as we're primarily focused on the Charmed HPC candidate release.

NucciTheBoss:
@dsloanm @jedel1043 this PR is read for a flesh and bone review :meat_on_bone: 

While Co-Pilot did have a valid point about `SnapError` being generic to all `snap` issues, and we're potentially masking underlying issues with the `snapd` subsystem, they're a natural flaw of subprocess wrappers. It's hard to provide granular exceptions when the wrapped command uses generic exit codes. E.g. all failed commands return exit code 1 regardless of the conditions that caused the failure.

I think there's room for future optimization here, but I think this is out of scope for introducing a new `exporters` module.

src/charmed_hpc_libs/ops/machine/snap.py
```diff
@@ -99,3 +182,12 @@ def is_active(self) -> bool:
...
         return "inactive" not in services[self._service]
+
+
+class SnapLifecycleManager(SnapOperationsManager, SnapServiceManager):
+    """Manage the full lifecycle operations of a snapped application."""
+
+    def __init__(self, snap: str, /, service: str = "") -> None:
+        # Call `__init__` explicitly to avoid opaque MRO resolution order.
+        SnapOperationsManager.__init__(self, snap)
+        SnapServiceManager.__init__(self, service or snap, snap=snap if service else None)
```
jedel1043: Not sure why we need this. All methods from `SnapOperationsManager` and `SnapServiceManager` are different, so there is no ambiguity on method resolution, right?

src/charmed_hpc_libs/ops/exporters.py
jedel1043: IMO property accessors should be "cheap", but calling snap -> snapd -> node-exporter seems like an expensive operation, so it might be better to represent these as proper methods.

jedel1043 (changes requested):
Looks good! I have a couple of questions and comments.

src/charmed_hpc_libs/ops/machine/snap.py
NucciTheBoss: Nah, it's because the order of arguments to the parents' `__init__` method is different. `SnapOperationsManager` takes the snap name as the first argument whereas `SnapServiceManager` takes service name as the first argument. `super()` would clobber this/not take into account the difference of argument ordering.

We could update the order of arguments for `SnapServiceManager`, but that would break compatibility for the existing usages of `SnapServiceManager` which isn't a change I'm willing to do right now 😅

I'll update the comment to be explicit that we call `__init__` directly instead of `super()` because the order of arguments is different for the parent class's constructors.

src/charmed_hpc_libs/ops/exporters.py
NucciTheBoss: Yeh, I was originally on the fence on whether to use properties or methods initially - properties just felt right at first since they're often used to change object configurations - but I agree with your comment that they should be "cheap".

Given that updating the snap configuration will actually restart the `node-exporter`, I agree that these should be changed to proper methods instead.

src/charmed_hpc_libs/ops/machine/snap.py
jedel1043: Hmmm kinda on the fence about hacking around inheritance then. Maybe a compositional approach would be better? We already do something like that for `slurm_ops`, which avoids depending on the initialization order of parent classes.

src/charmed_hpc_libs/ops/exporters.py
NucciTheBoss: These properties are now methods in the latest changes. Marking this as resolved.

src/charmed_hpc_libs/ops/machine/snap.py
NucciTheBoss: I wouldn't call it a "hack", but I understand the reservations. We're explicitly calling the parent classes instead of relying on the magic of the `super()`. There are trade-offs like we can still use an inheritance structure by explicitly calling the parent class's constructors, it's pretty brittle if we introduce another class we want `SnapLifecycleManager` to inherit from. This would be painful if there's potential conflicts that we have to resolve, but I don't think `SnapLifecycleManager` would get THAT complex. However, the brittleness means that this class won't scale well if it needs to become more complex in the future.

Therefore, since there's precedent for us to use composition instead of inheritance because that's what we already do in `slurm-ops`, I'll switch to this mechanism instead.

NucciTheBoss: `SnapLifecycleManager` is now built using composition instead of inheritance.

NucciTheBoss:
@jedel1043 ready for review round 2.

these new commits change properties to methods, and use composition instead of inheritance to construct the `SnapLifecycleManager` class. The composition is a bit more verbose code-wise, but it gets away from the brittleness of explicit calling parent class `__init__` methods.

Let me know if you have any questions!

src/charmed_hpc_libs/ops/machine/snap.py
```diff
@@ -48,7 +56,96 @@ def snap(*args: str, **kwargs: Any) -> tuple[str, int]:  # noqa D417
     return result.stdout, result.returncode
 
 
-class SnapServiceManager(ServiceManager):
+class _SnapManager:
+    """Manage snap packages."""
+
+    def __init__(self, snap: str) -> None:
+        self._snap = snap
```
jedel1043: Hmmm wondering if this is useful in general. It introduces double inheritance and makes the code a bit harder to read for just deduplicating at most 3 lines of code in 2 places. Maybe it would be better to have proper constructors on the managers instead?

jedel1043 (changes requested):

src/charmed_hpc_libs/ops/machine/snap.py
NucciTheBoss: It's useful in the sense that it's "DRY", however, from our chat yesterday we both agreed that this class is a gray-area since it's (a) super tiny and (b) I was worried about the name causing confusion for the unenlightened.

While it MIGHT feel wrong to the OOP elitists, I think since this code is only two lines we should just go with simplicity instead of cleverness and just copy + paste the `__init__` constructor where we need it.

NucciTheBoss: `_SnapManager` has been removed.

NucciTheBoss:
@jedel1043 this PR should be good to go here. I'll directly push a release commit to `main` after we merge this PR. Might be nice to use the [`release-please`](https://github.com/googleapis/release-please) workflow from Google to automatically push new releases after a PR is merged.

jedel1043 (approved):
LGTM! Nice work!

jedel1043:
BTW do you wanna clean up the commits a bit before merging? Since I see a lot of them are either bug patches or review fixes

NucciTheBoss:
I can just squash merge. I enabled it for this repository a while back after we had another PR that had a lot of minuscule commits.

PR #149, opened 2026-07-16
feat: add databag merging functionality to `_save_integration_data` method

### LINKED ISSUE #113: Merge data into integration databags rather than overwrite
Data class attributes are initialized to "zero" values on creation, e.g. `ControllerData` sets [`auth_key_id: str = ""`](https://github.com/charmed-hpc/hpc-libs/blob/main/src/hpc_libs/interfaces/slurm/common.py#L88), resulting in [`integration.save`](https://github.com/charmed-hpc/hpc-libs/blob/main/src/hpc_libs/interfaces/base.py#L293) calls overwriting data in the integration with these zero values. This creates a challenge when attempting to update only some of the values in a relation.

For example, in the [`slurmctld` charm](https://github.com/charmed-hpc/slurm-charms/tree/main/charms/slurmctld), we would like to update only the list of controllers in the `sackd` relation when `slurmctld` units join or leave a high availability (HA) setup. Attempting to do this with:

```python
self.sackd.set_controller_data(
    ControllerData(
        controllers=new_controllers,
    )
)
```

results in the `auth_key_id` and other attributes being overwritten with their zero values.

To solve this, the [HA PR](https://github.com/charmed-hpc/slurm-charms/pull/120) implements a [`_merge_controller_data`](https://github.com/charmed-hpc/slurm-charms/blob/5ff555ba62763a132f0804905c94cf0a2ab2e788/charms/slurmctld/src/charm.py#L555) function that reads existing data from the integration and updates only the controller list.

This solution works however it could be improved through support for merging from `hpc-libs` itself.

NucciTheBoss:
# Pre-submission checklist

## Summary of changes

This pull request adds databag merging functionality to the `_save_integration_data` method in the `Interface` base class.

A `merge: bool` parameter can be used now to direct the `_save_integration_data` method to either overwrite or merge integration data. If a dataclass with "advanced" objects is used, and `decoder` must also be passed to `_save_integration_data` so that the integration data can be correctly loaded. This behavior can be properly encapsulated by a downstream interface like `SlurmctldProvider` or `SlurmctldRequirer`.

- Closes https://github.com/canonical/charmed-hpc-libs/issues/113

## Docs

* [ ] I have created a pull request to add or update relevant documentation in [canonical/charmed-hpc-docs](https://github.com/canonical/charmed-hpc-docs) or another documentation location.

[//]: # (If documentation has been updated or added in a location other than canonical/charmed-hpc-docs, please note the location here.)

Or:

This is a non-user facing change to our HPC charm development SDK.

### DIFF
--- a/src/charmed_hpc_libs/interfaces/interface.py
+++ b/src/charmed_hpc_libs/interfaces/interface.py
@@ -16,6 +16,7 @@
 
 __all__ = ["Interface"]
 
+import dataclasses
 from collections.abc import Callable, Iterable
 from typing import Any, Literal
 
@@ -217,6 +218,8 @@ class Interface(ops.Object):
         integration_id: int | None = None,
         *,
         encoder: Callable[[Any], str] | None = None,
+        merge: bool = False,
+        decoder: Callable[[str], Any] | None = None,
     ) -> None:
         """Save integration data.
 
@@ -225,13 +228,48 @@ class Interface(ops.Object):
             target: Location to save object.
             integration_id: ID of integration to update.
             encoder: Callable that will be used to encode each field.
+            merge:
+                Whether to merge data into the integration databag rather than
+                overwriting. When ``True``, only fields whose values differ from
+                their dataclass defaults are written; existing values for unset
+                fields are preserved.
+            decoder:
+                Callable used to decode existing databag data when merging.
+                Only used when ``merge`` is True.
         """
         integrations = self.integrations
         if integration_id is not None:
             integrations = [self.get_integration(integration_id)]
 
+        if not merge:
+            for integration in integrations:
+                integration.save(data, target, encoder=encoder)
+
+            return
+
+        cls = type(data)
+        if not dataclasses.is_dataclass(cls):
+            raise TypeError(f"`merge=True` requires a dataclass instance, got '{cls.__name__}'")
+
+        set_fields = {
+            f.name: getattr(data, f.name)
+            for f in dataclasses.fields(cls)
+            if not self._field_is_default(f, getattr(data, f.name))
+        }
+
         for integration in integrations:
-            integration.save(data, target, encoder=encoder)
+            existing = integration.load(cls, target, decoder=decoder)
+            merged = dataclasses.replace(existing, **set_fields)
+            integration.save(merged, target, encoder=encoder)
+
+    @staticmethod
+    def _field_is_default(field: dataclasses.Field, value: Any) -> bool:
+        """Check if a field value is equal to its declared default."""
+        if field.default is not dataclasses.MISSING:
+            return value == field.default
+        if field.default_factory is not dataclasses.MISSING:
+            return value == field.default_factory()
+        return False
 
     @staticmethod
     def _is_integration_active(integration: ops.Relation) -> bool:
--- a/tests/unit/interfaces/test_interface.py
+++ b/tests/unit/interfaces/test_interface.py
@@ -121,10 +121,10 @@ def with_id(request) -> bool:
 
 
 class TestInterface:
-    """Test the `Interface` base class."""
+    """Test the ``Interface`` base class."""
 
     def test_integrations(self, mock_charm, integration_exists) -> None:
-        """Test the `integrations` property."""
+        """Test the ``integrations`` property."""
         with mock_charm(
             mock_charm.on.update_status(),
             state=testing.State(relations={make_integration()} if integration_exists else {}),
@@ -134,7 +134,7 @@ class TestInterface:
         assert len(manager.charm.interface.integrations) == (1 if integration_exists else 0)
 
     def test_get_integration(self, mock_charm, integration_exists) -> None:
-        """Test the `get_integration` method."""
+        """Test the ``get_integration`` method."""
         with mock_charm(
             mock_charm.on.update_status(),
             state=testing.State(relations={make_integration(id=1)} if integration_exists else {}),
@@ -148,7 +148,7 @@ class TestInterface:
                 manager.charm.interface.get_integration()
 
     def test_is_joined(self, mock_charm, integration_exists) -> None:
-        """Test the `is_joined` method."""
+        """Test the ``is_joined`` method."""
         with mock_charm(
             mock_charm.on.update_status(),
             state=testing.State(relations={make_integration()} if integration_exists else {}),
@@ -168,7 +168,7 @@ class TestInterface:
     def test_is_ready(
         self, mock_charm, integration_exists, data_is_available, with_id, expected
     ) -> None:
-        """Test the `is_ready` method."""
+        """Test the ``is_ready`` method."""
         with mock_charm(
             mock_charm.on.update_status(),
             state=testing.State(
@@ -191,7 +191,7 @@ class TestInterface:
         ),
     )
     def test_load_integration_data(self, mock_charm, with_id, target) -> None:
-        """Test loading application data with the `_load_integration_data` method."""
+        """Test loading application data with the ``_load_integration_data`` method."""
         with mock_charm(
             mock_charm.on.update_status(),
             state=testing.State(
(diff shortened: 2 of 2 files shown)


src/charmed_hpc_libs/interfaces/interface.py
```diff
@@ -225,13 +228,48 @@ def _save_integration_data(
...
+
+        cls = type(data)
+        if not dataclasses.is_dataclass(cls):
+            raise TypeError(f"`merge=True` requires a dataclass instance, got '{cls.__name__}'")
+
+        set_fields = {
+            f.name: getattr(data, f.name)
+            for f in dataclasses.fields(cls)
+            if not self._field_is_default(f, getattr(data, f.name))
+        }
```
dsloanm, line 283: Does this prevent fields from ever being reset to default?

For example, if we had a default `some_field: str = ""` and wanted to set it to `"some value"` then back to `""`, we wouldn't be able to?

tests/unit/interfaces/test_interface.py
```diff
@@ -121,10 +121,10 @@ def with_id(request) -> bool:
 
 
 class TestInterface:
-    """Test the `Interface` base class."""
+    """Test the ``Interface`` base class."""
```
dsloanm, line 122: Was the change from ` to `` in docstrings intentional in this file?

dsloanm (commented):
This is looking good and will be highly useful once merged. I had a couple of questions before approval

src/charmed_hpc_libs/interfaces/interface.py
NucciTheBoss: Good question :eyes: - I'll look into this in further detail later today. It likely won't let you purposely set the default if `merge=True`. I'll see what can be done here.

tests/unit/interfaces/test_interface.py
NucciTheBoss: Yes. The tl;dr is that the double backticks are required for code syntax highlighting to be displayed in docstring overviews whereas the single backticks don't enable code syntax highlighting.

I learned this a couple weeks back when reviewing some code, so I've been slowly rolling out the change from single to double backticks in my PRs since. Not really a high priority change though so that's why I haven't hardcore pushed for us to do a bulk update of all our docstrings to use double instead of single backticks for code syntax highlighting.

src/charmed_hpc_libs/interfaces/interface.py
NucciTheBoss: Alrighty - the latest commit addresses this situation: https://github.com/canonical/charmed-hpc-libs/pull/149/commits/87b9705325b63de0691db93db7ba66276118c350

Your observation was correct that the original implementation of the `merge` parameter would've prevented you from resetting the value of a field back to its default even if you intentionally set the value of field back to its default. To fix this, I introduced a new `reset` parameter that accepts a set of field names that you want to reset while merging new data into the integration databag.

It may appear clunky at first, but this is the best approach I found that wasn't too tightly integrated with the implementation of Python's dataclasses or Pydantic's data models. This implementation is also interoperable with native dataclasses or pydantic dataclasses.

NucciTheBoss:
@dsloanm I updated my PR so that you now have the ability to reset specific fields in an integration databag when merging dataclasses together. Ready for review round 2 electric boogalo :sunglasses:

src/charmed_hpc_libs/interfaces/interface.py
dsloanm: This works but we may want to rethink the partially-initialized object approach we're using since it's necessitating these workarounds. Not in this PR but potentially the next charmed-hpc-libs major version.

src/charmed_hpc_libs/interfaces/interface.py
```diff
@@ -225,13 +229,75 @@ def _save_integration_data(
...
+        set_fields = {
+            field.name: getattr(data, field.name)
+            for field in dataclasses.fields(cls)
+            if not self._field_is_default(field, getattr(data, field.name))
+        }
+        reset_fields = {
+            field.name: self._get_field_default(field)
+            for field in dataclasses.fields(cls)
+            if field.name in reset
+        }
```
dsloanm, line 283: It looks like unknown field names in `reset` are silently ignored, so a typo like `reset={"controller"}` (missing the "s") for [`ControllerData`](https://github.com/canonical/slurm-charms/blob/cde79ad491a86ffce0f53e36e2e87d50652ec00c/pkg/slurmctld-interface/src/charmed_slurm_slurmctld_interface/__init__.py#L64) would silently fail to reset `controllers`.

So we're not caught out by this, it would be useful to add validation that fields in `reset` correspond to an actual field, such as (courtesy of GLM):

```python
field_names = {f.name for f in dataclasses.fields(cls)}
unknown = reset - field_names
if unknown:
    raise ValueError(f"unknown field(s) in `reset`: {sorted(unknown)}")
```

src/charmed_hpc_libs/interfaces/interface.py
```diff
@@ -225,13 +229,75 @@ def _save_integration_data(
...
+
         for integration in integrations:
-            integration.save(data, target, encoder=encoder)
+            existing = integration.load(cls, target, decoder=decoder)
+            merged = dataclasses.replace(existing, **(set_fields | reset_fields))
```
dsloanm, line 286: Would be good to add to the docstring that `reset_fields` takes priority over `set_fields`, or change the code to raise an error if a field is in both.

 Currently it's not clear to a caller what happens if they pass `data = ExampleData(name="foo")` and `reset={"name"}`

dsloanm (changes requested):
Some suggested changes then this should be good to go.

src/charmed_hpc_libs/interfaces/interface.py
NucciTheBoss: Yeh, I agree that it's not the most _ideal_ fix, but it gets the job done :sweat_smile: 

I was also looking into the sentinel value approach where the base value effectively tells you whether or not a value was modified, but this seems tightly integrated with the object's type such as whether it's a native dataclass or a pydantic dataclass.

I agree with your comment that we should just rip off the band-aid and mandate that objects modelling integration data must be a pydantic dataclass, but I think that's for when we formally publish a v1 API of `charmed-hpc-libs`. Probably something to visit once we publish the first stable release of Charmed HPC.

All that said, I'm going to resolve this convo since we've identified that this is something that this is something that needs to be addressed when we formalize the API of `charmed-hpc-libs`.

src/charmed_hpc_libs/interfaces/interface.py
NucciTheBoss: Good idea! Best to avoid silent errors/raise an exception if the charm author isn't actually resetting the value the intend to reset.

NucciTheBoss: Additional runtime inspection added in https://github.com/canonical/charmed-hpc-libs/pull/149/commits/017e2826df2e06dbf57b32901ac90a52e379e6a0.

src/charmed_hpc_libs/interfaces/interface.py
NucciTheBoss: Added statement about precedence to the description of the `reset` parameter in the method docstring in https://github.com/canonical/charmed-hpc-libs/pull/149/commits/017e2826df2e06dbf57b32901ac90a52e379e6a0.

NucciTheBoss:
Alrighty @dsloanm, addressed your new comments from round 2. Ready for review again :mag:

dsloanm (approved):
LGTM!

NucciTheBoss:
> LGTM!

Awesome! I'm going to push a commit that does a minor version bump and then merge this PR :sunglasses: 

Edit: I had to do a force push to this branch since it turned out that my development worktree was relatively stale. _This wasn't the first time that I took a stab at this issue..._

